PR #13, opened 2018-06-08
Update localdate properties to Instant

meletios:


meletios:
will be handled later

PR #20, opened 2018-07-25
map INACTIVE product status to the TERMINATED product state and map P…

meletios:
…ENDING status to everything else apart the ACTIVE cases


johantjellden:
We will rewrite this in another way.

PR #22, opened 2018-07-31
Bugfix/app status descriptions

johantjellden:


src/main/java/com/hedvig/productPricing/service/web/dto/InsuranceDTO.java
```diff
@@ -66,7 +66,9 @@ public InsuranceDTO(Clock clock, ProductEntity p, List<PerilEntity> perils, User
       this.addressStreet = ue.address.getStreet();
     }
 
-    this.status = ProductStatus.createStatus(clock, p.state, p.activeFrom, p.activeTo);
+      final ProductStatus.Status status = ProductStatus.createStatus(clock, p.state, p.activeFrom, p.activeTo);
```
hedvigoscar: formatting seems off here - I suggest we enable google style in editors :+1:

PR #31, opened 2018-08-07
add new endpoint to query insurance by their activation dates, add me…

meletios:
…thod to productrepository, add new DTO to return inuranceId, memberId and activationDate


src/main/java/com/hedvig/productPricing/service/query/ProductRepository.java
```diff
@@ -1,51 +1,62 @@
...
+
+  @Query(
+      " from ProductEntity p"
+          + " where p.activeFrom is not null"
+          + " and date_part ('month', p.activeFrom ) <= :month"
+          + " and date_part ('year', p.activeFrom ) <= :year"
+          + " and p.member.id = :memberId")
+  List<ProductEntity> findBillingByDateByMemberId(
+      @Param("month") int month, @Param("year") int year, @Param("memberId") String memberId);
+
```
meletios: below you can find the new method for searching by activationDate, the rest are formatting changes :/

PR #34, opened 2018-08-15
using the insuranceId in order to update set or update the activation…

meletios:
… date instead of the memberId


src/test/java/com/hedvig/productPricing/service/web/InternalControllerTest.java
```diff
@@ -36,6 +38,8 @@
 @TestPropertySource(properties = "hedvig.product-pricing.files.location = src/test/resources/")
 public class InternalControllerTest {
 
+    private final static UUID ONE_INSURANCE_ID = UUID.randomUUID();
```
johantjellden: Generate a static UUID to use here. There is a *uuid* tool you can use from terminal.

src/main/java/com/hedvig/productPricing/service/commands/ActivateInsuranceAtDate.java
```diff
@@ -1,14 +1,14 @@
...
 
+import java.time.LocalDate;
+import java.util.UUID;
 import lombok.Value;
 import org.axonframework.commandhandling.TargetAggregateIdentifier;
 
-import java.time.LocalDate;
-
 @Value
 public class ActivateInsuranceAtDate {
```
johantjellden: Rename class to follow command style.

PR #38, opened 2018-09-11
Do not charge members in state quote

johantjellden:


src/main/java/com/hedvig/productPricing/service/service/InsuranceBillingServiceImpl.java
```diff
@@ -213,7 +212,7 @@ public PricingResult getPricingResult(
     List<ProductEntity> billingList =
         productRepository.findBillingByDate(period.getMonthValue(), period.getYear());
 
-    return calcualteBilling(period, billingList)
+    return calculateBilling(period, billingList)
```
meletios: :tada:

PR #39, opened 2018-09-13
Feature/bordereau

meletios:


src/main/java/com/hedvig/productPricing/service/service/InsuranceBillingServiceImpl.java
```diff
@@ -34,231 +35,262 @@
...
+            .orElse(new InsuranceBillingDTO(memberId, Money.of(0L, "SEK")));
+    }
+
+    @Override
+    public List<InsuranceBordereauDTO> getMonthlyBordereau(YearMonth period,
+        ProductTypes productType) {
+        List<ProductEntity> productList = productRepository
+            .findBillingByDate(period.getMonthValue(), period.getYear());
+
+        Stream<InsuranceBillingInfo> filterdList = productList.stream()
```
hedvigoscar: ^filterd^filtered

src/main/java/com/hedvig/productPricing/service/service/InsuranceBillingServiceImpl.java
```diff
@@ -34,231 +35,262 @@
...
+            .filter(x -> (x.state == ProductStates.SIGNED || x.state == ProductStates.TERMINATED)
```
hedvigoscar: Would it be wise to use `x.state != ProductStates.QUOTE`?

Either way we may run into trouble if we introduce more `ProductState`s. Perhaps a function `boolean isSignedProductState(ProductState state)`?

meletios: Correct, a method feels better. Done!

src/main/java/com/hedvig/productPricing/service/service/InsuranceBillingServiceImpl.java
meletios: fixed

PR #45, opened 2018-10-08
Feature/redis publish

meletios:


src/main/java/com/hedvig/productPricing/service/service/ProductCreatedRedisMessage.java
```diff
@@ -0,0 +1,12 @@
+package com.hedvig.productPricing.service.service;
+
+
+import lombok.AllArgsConstructor;
+import lombok.Value;
+
+@Value
+@AllArgsConstructor
+public class ProductCreatedRedisMessage {
```
hedvigoscar: Could we send this kind of object instead?:
```
{
  "offer": {
    "status": "SUCCESS/FAIL"
  }
}
```

It would surprisingly make it a lot easier for me 🙁

hedvigoscar: Also `@AllArgsConstructor` is not needed when using `@Value`, it autogenerates it :wink:

meletios: of course :+1:

PR #51, opened 2018-10-15
Migrate to spring boot 2

johantjellden:


src/test/java/com/hedvig/productPricing/ProductPricingApplicationTests.java
```diff
@@ -15,181 +14,179 @@
...
-	public void contextLoads() {
-		given().
-				header("hedvig.token", "1337").
-		when().
-			get("/insurance").
-		then().
-			statusCode(200).
-			body("status", equalTo("PENDING"));
-	}
+  @Test
```
meletios: Should we delete them?

PR #85, opened 2018-10-30
Validate underwriting limits

hedvigoscar:


src/main/java/com/hedvig/productPricing/service/web/ProductPricingController.java
```diff
@@ -289,6 +291,10 @@ public ProductPricingController(
 
         List<PerilDTO> perils = this.insuranceBillingService.getPerils(request.getProductType());
 
+        if (!validateProductTypeUnderwritingLimits(request)) {
+          return ResponseEntity.status(HttpStatus.FORBIDDEN).build();
```
meletios: 400 feels better than a 403 since the client didn't send the correct data.

src/main/java/com/hedvig/productPricing/service/web/ProductPricingController.java
```diff
@@ -325,6 +331,19 @@ public ProductPricingController(
         return ResponseEntity.ok().body(String.format("{\"id\":\"%s\"}", id));
     }
 
+    private boolean validateProductTypeUnderwritingLimits(WebOnboardingQuoteRequest request) {
```
meletios: No need to be fixed now, but can we use it, when we create the product from the App or ModifyInsurance endpoints?

PR #86, opened 2018-10-31
Feature/signing mail

meletios:


src/main/java/com/hedvig/productPricing/service/service/InsuranceNotificationServiceImpl.java
```diff
@@ -16,30 +16,39 @@
...
-      String memberId, Boolean insuredAtOtherCompany, InsuranceCompaniesSE currentInsurer) {
-    if (currentInsurer.equals(InsuranceCompaniesSE.OTHER)) {
-      notifyUnknownInsurer(memberId);
+  public void sendInsuranceSignedNotification(
+    String memberId,
+    Boolean insuredAtOtherCompany,
+    InsuranceCompaniesSE currentInsurer) {
+
+    if (!insuredAtOtherCompany) {
+      notificationService.sendInsuranceSignedEmail(memberId, false, null);
```
johantjellden: We should tell notificaiton service that this has happend and let notification decide if we send an email. So rename it to something like **insurancesSigned**.

src/main/java/com/hedvig/productPricing/service/serviceIntegration/notificationService/NotificationServiceImpl.java
```diff
@@ -0,0 +1,34 @@
...
+  private NotificationServiceClient notificationServiceClient;
+
+  public NotificationServiceImpl(NotificationServiceClient notificationServiceClient) {
+    this.notificationServiceClient = notificationServiceClient;
+  }
+
+  @Override
+  public void sendInsuranceSignedEmail(
+    String memberId,
+    boolean switchingFromCurrentInsurer,
```
johantjellden: Maybe this boolean is reduntant, we either have a currentInsurer or not.

src/main/java/com/hedvig/productPricing/service/sagas/ContractSignedSaga.java
```diff
@@ -3,38 +3,53 @@
...
+    val productEntity = productRepository.findById(event.getProductId());
+
+    if (productEntity.isPresent()) {
+      ProductEntity product = productEntity.get();
+
+      InsuranceCompaniesSE insurer = InsuranceCompaniesSE.create(product.currentInsurer);
+      try {
+        insuranceNotificationService.sendInsuranceSignedNotification(
+          event.getMemberId(),
+          product.insuredAtOtherCompany && product.currentInsurer != null,
```
johantjellden: I think that we should treat **product.insuredAtOtherCompany == true ====> product.currentInsurer != null** as an invariant that cannot be broken. So maybe we should abort here and log an error, the point is that we don't want to send a welcome you insurance is active now to someone that has a existing insurance.

src/main/java/com/hedvig/productPricing/service/service/InsuranceNotificationServiceImpl.java
meletios: :+1:

src/main/java/com/hedvig/productPricing/service/serviceIntegration/notificationService/NotificationServiceImpl.java
meletios: i have change name, i agree it was too confusing

src/main/java/com/hedvig/productPricing/service/sagas/ContractSignedSaga.java
meletios: added a check

PR #88, opened 2018-11-02
Feature/signing mail

meletios:


src/main/java/com/hedvig/productPricing/service/sagas/ContractSignedSaga.java
```diff
@@ -40,11 +40,15 @@ public void start(ContractSignedEvent event) {
     if (productEntity.isPresent()) {
       ProductEntity product = productEntity.get();
       try {
-        if (product.insuredAtOtherCompany && product.currentInsurer == null){
-          throw new RuntimeException("current insurer not found");
+
+        if ((product.insuredAtOtherCompany && product.currentInsurer == null)
```
hedvigoscar: Should this perhaps be validated at a higher level?

PR #89, opened 2018-11-02
Proof of Concept: Use Jib to build better Docker images

hedvigoscar:
This is a proof of concept and should probably not be merged yet.


FredrikAreschoug:
It was a nice poc but I will close it now :smile:

PR #138, opened 2019-03-13
WIP change start date

hedvigoscar:
This is not really done yet but I'm opening a PR so that we can start programming against this interface


src/main/java/com/hedvig/productPricing/service/aggregates/MemberAggregate.java
```diff
@@ -179,16 +181,34 @@ public void modifyProduct(ModifyProductCommand cmd) {
     apply(new CertificateGenerationIssuedEvent(cmd.getInsuranceIdToReplace()));
   }
 
+  @CommandHandler
+  void requestStartDate(RequestStartDateCommand cmd) {
+    Optional<Product> maybeProduct = products
+      .stream()
+      .filter(x -> x.state == ProductStates.QUOTE)
```
palmfjord: no `x` plz :/

src/main/java/com/hedvig/productPricing/service/aggregates/MemberAggregate.java
```diff
@@ -179,16 +181,34 @@ public void modifyProduct(ModifyProductCommand cmd) {
...
+      .findFirst();
+
+    if (!maybeProduct.isPresent()) {
+      throw new RuntimeException("No valid candidate products to set a custom start date for");
```
palmfjord: maybe we throw in some more data? Also should this maybe be some kind of `IllegalStateException` or so?

src/main/java/com/hedvig/productPricing/service/aggregates/MemberAggregate.java
```diff
@@ -202,10 +222,20 @@ void signedContract(SingedContractCommand cmd, PdfCreator pdfCreator) throws IOE
                   .reduce((a, b) -> b); // Take latest if there are multiple products
 
       if (productToBeSigned.isPresent()) {
+        Product product = productToBeSigned.get();
+        LocalDateTime activeFrom;
+        if (product.requestedStartDate != null) {
+          activeFrom = product.requestedStartDate;
+        } else if (this.currentInsurer == null) {
+          activeFrom = LocalDateTime.ofInstant(Instant.now(), ZoneOffset.UTC);
```
palmfjord: shouldn't this technically be `Europe/Stockholm`?

palmfjord: ping @meletios

src/main/java/com/hedvig/productPricing/service/events/StartDateRequestedEvent.java
```diff
@@ -0,0 +1,13 @@
+package com.hedvig.productPricing.service.events;
+
+import java.time.LocalDateTime;
+import java.util.UUID;
+
+import lombok.Value;
+
+@Value
+public class StartDateRequestedEvent {
+  String memberId;
```
palmfjord: we should specify visibility for this, preferably `private` imo

src/main/java/com/hedvig/productPricing/service/commands/RequestStartDateCommand.java
```diff
@@ -0,0 +1,14 @@
...
+import java.time.LocalDateTime;
+
+import org.axonframework.commandhandling.TargetAggregateIdentifier;
+
+import lombok.Value;
+
+@Value
+public class RequestStartDateCommand {
+  @TargetAggregateIdentifier
+  String memberId;
```
palmfjord: same, visibility plz

src/main/java/com/hedvig/productPricing/service/web/ProductPricingController.java
```diff
@@ -837,6 +840,34 @@ private void createOrUpdatePeril(PerilDTO p) {
...
+    val requestedStartDate = dto.getRequestedStartDate();
+    if (requestedStartDate == null) {
+      commandBus.sendAndWait(
+        new RequestStartDateCommand(
+          memberId,
+          requestedStartDate
+        )
+      );
+    }
+    if (requestedStartDate.isBefore(LocalDateTime.now())) {
```
palmfjord: timezone plz :D

src/main/java/com/hedvig/productPricing/service/aggregates/MemberAggregate.java
meletios: Yes it should be Europe/Stockholm

meletios: Instead of passing the activation date in `RequestStartDateCommand` and then to the conractSignedCommand, we should instead pass it with the `SignedContractCommand`. This will save us time on mapping the events for reporting and it will make a lot more simpler.

src/main/java/com/hedvig/productPricing/service/aggregates/MemberAggregate.java
```diff
@@ -179,16 +181,34 @@ public void modifyProduct(ModifyProductCommand cmd) {
     apply(new CertificateGenerationIssuedEvent(cmd.getInsuranceIdToReplace()));
   }
 
+  @CommandHandler
+  void requestStartDate(RequestStartDateCommand cmd) {
```
meletios: Check comment on line 230

src/main/java/com/hedvig/productPricing/service/aggregates/MemberAggregate.java
hedvigoscar: I'm not opposed to this idea, but there are some problems surrounding it: namely where to store the start date (should we just keep it in memory in the app? in some form of disk storage?).

There is also the problem of passing this data down to bot-service, member-service or whichever service handles signs today and then passing it along to product-pricing, which complicates the feature slightly.

src/main/java/com/hedvig/productPricing/service/events/StartDateRequestedEvent.java
hedvigoscar: https://projectlombok.org/features/Value I think that the lombok value annotation actually does this by default 🙂

src/main/java/com/hedvig/productPricing/service/web/ProductPricingController.java
hedvigoscar: Should I accept a different type of date here?

hedvigoscar:
I feel like I'm getting less and less qualified to make changes to product-pricing with all the new time zone things and time you have invested in this service so I'm not sure that I'm actually qualified to continue this work 😞

src/main/java/com/hedvig/productPricing/service/events/StartDateRequestedEvent.java
palmfjord: W0o0o0ooo0ot O.o

src/main/java/com/hedvig/productPricing/service/web/ProductPricingController.java
palmfjord: no but `LocalDateTime.now()` defaults to the system timezone and we shouldn't depend on that so it would be better with something like `LocalDateTime.now(ZoneId.of("Europe/Stockholm"))`

src/main/java/com/hedvig/productPricing/service/events/StartDateRequestedEvent.java
hedvigoscar: Hax 😉

palmfjord:
@meletios 
> we should instead pass it with the SignedContractCommand. This will save us time on mapping the events

i don't understand this, don't we need to change all `signedcontractevent`s anyways? meaning we have to reset and replay everything - or make yet another event listener. Both feel quite sucky tbh hehe

johantjellden:
:(

PR #143, opened 2019-03-21
Feature/awesome mi report nighttime coding

meletios:


src/main/java/com/hedvig/productPricing/service/policies/PolicyEventListener.java
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  @EventHandler
+  public void on(final ContractSignedEvent e, final @Timestamp Instant timestamp) {
+    if (BLACKLISTED_MEMBERS.contains(e.getMemberId())) {
+      return;
+    }
+
+    if (e.getActiveFrom() == null) {
+      log.info("Received contract signed event without activation date, skipping");
```
palmfjord: i think logging here should include some member ids and stuff so its easier to track down if things go wrong

src/main/java/com/hedvig/productPricing/service/service/reports/EarnedGrossWrittenPremiumGenerator.java
```diff
@@ -0,0 +1,51 @@
...
+        underwritingYear.atMonth(12).atEndOfMonth(),
+        timeOfKnowledge
+      ).stream(),
+      timeOfKnowledge,
+      underwritingYear,
+      lineOfBusiness
+    );
+
+    return knownPolicyTermPeriodsInLineOfBusiness.stream()
+      .filter(policyTermPeriod1 -> !YearMonth.from(policyTermPeriod1.getPeriodStartInStockholm()).isAfter(period))
```
palmfjord: `1`??

src/main/java/com/hedvig/productPricing/service/service/reports/LineOfBusiness.java
```diff
@@ -0,0 +1,24 @@
...
+      case BRF:
+      case RENT:
+      case SUBLET_BRF:
+      case SUBLET_RENTAL:
+        return HOUSEHOLD;
+      case STUDENT_BRF:
+      case STUDENT_RENT:
+        return HOUSEHOLD_STUDENT;
+      default:
+        throw new RuntimeException("Cannot convert ProductType " + t + " to LOB");
```
palmfjord: illlegalargument?

src/main/java/com/hedvig/productPricing/service/service/reports/PolicyCountGenerator.java
```diff
@@ -0,0 +1,47 @@
...
+  public PolicyCountGenerator(final PolicyRepository policyRepository) {
+    this.policyRepository = policyRepository;
+  }
+
+  public int generate(final Year underwritingYear, final YearMonth period, final LineOfBusiness lineOfBusiness) {
+    final Instant timeOfKnowledge = period.atEndOfMonth().atTime(23, 59, 59, 999_999_999)
+      .atZone(ZoneId.of("Europe/Stockholm"))
+      .toInstant();
+
+    return findKnownPolicyTermPeriodsInLineOfBusiness(
```
palmfjord: not sure this is 100% correct, what if the policy for this ToK doesn't have any periods ie it has been terminated before it was activated with inception date was locked? then it won't show up right?

palmfjord: what would be more correct i guess is to filter out policies within this underwriting year with ToK > now

src/main/java/com/hedvig/productPricing/service/service/reports/MiReportGenerator.java
```diff
@@ -0,0 +1,85 @@
+package com.hedvig.productPricing.service.service.reports;
```
palmfjord: maybe some logs messages here in case things go south?

PR #144, opened 2019-03-27
Feature/reporting guess insurance types

palmfjord:


src/main/java/com/hedvig/productPricing/service/policies/PolicyManagerImpl.java
```diff
@@ -184,4 +183,24 @@ public void cancelPolicy(final Policy policy, final Instant now, final LocalDate
       .filter(policy -> policy.getTermSnapshots(timestamp).size() > 0)
       .collect(toList());
   }
+
+  public List<Policy> findPoliciesFor(final List<String> memberIds, final Instant timestamp) {
+    final List<List<String>> chunkedMemberIds = new ArrayList<>();
```
palmfjord: i think we should move this out to a separate method, i do not want to come back to this without understanding why we're doing this

src/main/java/com/hedvig/productPricing/service/web/InternalController.java
```diff
@@ -336,6 +292,22 @@ public ProductSearchResultDTO searchPaged(ProductSearchRequestDTO searchRequest)
         return ok(insurances);
     }
 
+      @PostMapping(path = "/policies/guess-types/{period}")
```
meletios: Should this be under the report controller?

PR #151, opened 2019-04-09
Feature/update customer io fields

meletios:


src/main/java/com/hedvig/productPricing/service/service/segmentExporter/EventListener.java
```diff
@@ -31,80 +34,93 @@
...
-                ? InsuranceCompaniesSE.create(e.getCurrentInsurer()).formattedName()
-                : null,
-            source == null ? SignupSource.APP : source);
+      new ProductCreatedTrait(
+        e.getInsuredAtOtherCompany(),
+        e.getInsuredAtOtherCompany()
+          ? InsuranceCompaniesSE.create(e.getCurrentInsurer()).formattedName()
+          : null,
+        source == null ? SignupSource.APP : source,
+        e.getAddress() != null ? e.getAddress().getStreet() : null,
```
palmfjord: can we maybe move this into a util function or varible or something? feels tedious

src/main/java/com/hedvig/productPricing/service/service/segmentExporter/EventListener.java
```diff
@@ -31,80 +34,93 @@
...
+        e.getAddress() != null ? e.getAddress().getCity() : null,
+        e.getAddress() != null ? e.getAddress().getZipCode() : null,
+        e.getPersonsInHouseHold(),
+        e.getCurrentTotalPrice(),
+        e.getIsStudentInsurance(),
+        timestamp.atZone(ZoneId.of("Europe/Stockholm"))
```
palmfjord: isn't this timestamp technically UTC? And doen't CIO expect timestamps as UTC?

src/main/java/com/hedvig/productPricing/service/service/segmentExporter/EventListener.java
```diff
@@ -113,43 +129,51 @@ public void on(MemberCreatedEvent evt) {
...
+        .map(eventMessage -> (ModifiedProductCreatedEvent) eventMessage.getPayload())
+        .filter(event -> event.getId().equals(e.getInsuranceIdToReplace()));
+
+    Optional<ModifiedProductCreatedEvent> modifiedProductCreatedEventMaybe = modifiedProductCreatedEventStream.findFirst();
+
+    if (modifiedProductCreatedEventMaybe.isPresent()) {
+      ModifiedProductCreatedEvent modifiedProductCreatedEvent = modifiedProductCreatedEventMaybe.get();
+
+      ProductModifiedTrait trait = new ProductModifiedTrait(
+        modifiedProductCreatedEvent.getAddress() != null ? modifiedProductCreatedEvent.getAddress().getStreet() : null,
```
palmfjord: again? 😬

src/main/java/com/hedvig/productPricing/service/service/segmentExporter/InsuranceDataTrait.java
```diff
@@ -0,0 +1,21 @@
+package com.hedvig.productPricing.service.service.segmentExporter;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Value;
+
+import java.math.BigDecimal;
+
+@Value
+public class InsuranceDataTrait {
```
palmfjord: could we maybe use `@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)` instead of setting every field manually? (unsure if it works but maybe)

src/main/java/com/hedvig/productPricing/service/service/segmentExporter/SegmentEventListener.java
```diff
@@ -0,0 +1,104 @@
...
+    this.eventStore = eventStore;
+  }
+
+  @EventHandler
+  public void on(ProductCreatedEventV2 e, @Timestamp Instant timestamp) {
+    val traits = new InsuranceDataTrait(
+      e.getPersonsInHouseHold(),
+      e.getCurrentTotalPrice(),
+      e.getIsStudentInsurance(),
+      timestamp.atZone(ZoneId.of("Europe/Stockholm")).truncatedTo(DAYS).toEpochSecond()
```
palmfjord: again, are we sure these timestamps should be considered `europe/stockholm`?

src/main/java/com/hedvig/productPricing/service/service/segmentExporter/SegmentEventListener.java
```diff
@@ -0,0 +1,104 @@
...
+
+    Optional<ModifiedProductCreatedEvent> modifiedProductCreatedEventMaybe = modifiedProductCreatedEventStream.findFirst();
+
+    if (modifiedProductCreatedEventMaybe.isPresent()) {
+      ModifiedProductCreatedEvent modifiedProductCreatedEvent = modifiedProductCreatedEventMaybe.get();
+
+      val traits = new InsuranceDataTrait(
+        modifiedProductCreatedEvent.getPersonsInHouseHold(),
+        modifiedProductCreatedEvent.getCurrentTotalPrice(),
+        modifiedProductCreatedEvent.getIsStudentInsurance() != null ? modifiedProductCreatedEvent.getIsStudentInsurance() : false,
```
palmfjord: i love java

src/main/java/com/hedvig/productPricing/service/service/segmentExporter/InsuranceDataTrait.java
meletios: we cannot use this to all the cases, since we are using different names for the json property than the class property. ` @JsonProperty("**insurance_address_city**")
  @JsonInclude(value = JsonInclude.Include.NON_NULL)
  String **city**;
`

src/main/java/com/hedvig/productPricing/service/service/segmentExporter/SegmentEventListener.java
meletios: we are following the same strategy for the previous date related variables, so i think its better if we keep it that way.

src/main/java/com/hedvig/productPricing/service/service/segmentExporter/InsuranceDataTrait.java
palmfjord: But i mean it would help for quite a few right? 😬

src/test/java/com/hedvig/productPricing/service/service/segmentExporter/SegmentEventListenerTest.java
```diff
@@ -0,0 +1,175 @@
+package com.hedvig.productPricing.service.service.segmentExporter;
```
palmfjord: this ❤👇

PR #158, opened 2019-04-29
Feature/backfill member accounts

vonElfvin:


src/main/java/com/hedvig/productPricing/service/serviceIntegration/accountService/AccountService.java
```diff
@@ -0,0 +1,7 @@
+package com.hedvig.productPricing.service.serviceIntegration.accountService;
+
+import org.springframework.http.ResponseEntity;
+
+public interface AccountService {
+  ResponseEntity<?> createAccount(String memberId);
```
vonElfvin: Any suggestions?

Account-service returns "ResponseEntity<Void>

src/main/java/com/hedvig/productPricing/service/serviceIntegration/accountService/AccountServiceClient.java
```diff
@@ -0,0 +1,17 @@
+package com.hedvig.productPricing.service.serviceIntegration.accountService;
+
+import com.hedvig.productPricing.config.FeignConfiguration;
```
vonElfvin: Note the location of the FeignConfiguration, found similar usage in other services. TODO: Move to a central config folder?

src/main/java/com/hedvig/productPricing/service/serviceIntegration/accountService/AccountService.java
palmfjord: Does `ResponseEntity<Void>` not work here then?

src/main/java/com/hedvig/productPricing/service/accounts/AccountEventListener.java
```diff
@@ -0,0 +1,38 @@
...
+  public AccountEventListener(
+    AccountService accountService
+  ) {
+    this.accountService = accountService;
+  }
+
+  @EventHandler
+  public void on(ContractSignedEvent e) {
+    String memberId = e.getMemberId();
+    if (!BLACKLISTED_MEMBERS.contains(e.getMemberId())) {
```
palmenhq: i think we should unnegate this and make an early return to keep the indentation level a bit smaller :)

PR #159, opened 2019-05-06
init apple insurance

meletios:


src/main/java/com/hedvig/productPricing/service/web/InternalController.java
```diff
@@ -9,20 +9,19 @@
 import com.hedvig.productPricing.service.aggregates.Product.ProductTypes;
```
palmfjord: oh myy... this file. Can we plz reformat it separately? It's completely impossible to understand what's going on here with whitespace diff on, and i do not want to come back trying to use git blame when there's both reformatting and changes in the same pr

src/main/java/com/hedvig/productPricing/service/web/InternalController.java
```diff
@@ -343,262 +341,312 @@ public ProductSearchResultDTO searchPaged(ProductSearchRequestDTO searchRequest)
...
+    return ResponseEntity.ok(new ProductToSignStatusDTO(false, false));
+  }
 
-        HashMap<String, String> perilsHashMap = new HashMap<>();
-        for (PerilDTO p : perils) {
-            log.debug(p.id);
-            perilsHashMap.put(p.id, p.state);
-        }
+  @PostMapping("/initiateAppleProduct")
+  public ResponseEntity<?> initiateAppleProduct(@RequestBody AppleInitializationRequest req) {
```
palmfjord: `ResponseEntity<Void>` no?

src/main/java/com/hedvig/productPricing/service/web/InternalController.java
```diff
@@ -343,262 +341,312 @@ public ProductSearchResultDTO searchPaged(ProductSearchRequestDTO searchRequest)
...
+      )
+    );
+
+    commandBus.sendAndWait(
+      new SingedContractCommand(
+        req.getMemberId(),
+        "AppleUser",
+        "AppleUser",
+        "AppleUser",
+        Instant.now(),
```
palmfjord: maybe a fixed time?

src/main/java/com/hedvig/productPricing/service/web/dto/AppleInitializationRequest.java
```diff
@@ -0,0 +1,19 @@
+package com.hedvig.productPricing.service.web.dto;
+
+import lombok.Value;
+
+@Value
+public class AppleInitializationRequest {
+  private String memberId;
+  private String personalNumber;
```
palmfjord: Isn't it better to put things like this in the config so we can only make one and know who he is?

src/main/java/com/hedvig/productPricing/service/web/InternalController.java
```diff
@@ -343,262 +341,312 @@ public ProductSearchResultDTO searchPaged(ProductSearchRequestDTO searchRequest)
...
+        req.getLastName(),
+        timBirthDay,
+        false,
+        new com.hedvig.productPricing.service.serviceIntegration.botService.dto.Address(
+          req.getStreet(),
+          req.getCity(),
+          req.getZipCode(),
+          0
+        ),
+        Float.valueOf("100"),
```
palmfjord: why not just `100f`?

src/main/java/com/hedvig/productPricing/service/web/InternalController.java
```diff
@@ -343,262 +341,312 @@ public ProductSearchResultDTO searchPaged(ProductSearchRequestDTO searchRequest)
...
+        req.getMemberId(),
+        "AppleUser",
+        "AppleUser",
+        "AppleUser",
+        Instant.now(),
+        req.getPersonalNumber()
+      )
+    );
+
+    return ResponseEntity.ok().build();
```
palmfjord: `noContent()`?  😬

src/main/java/com/hedvig/productPricing/service/policies/PolicyEventListener.java
```diff
@@ -27,7 +28,11 @@
   private static final ZoneId STOCKHOLM = ZoneId.of("Europe/Stockholm");
   private final PolicyManager policyManager;
   private final EventStore eventStore;
-  private final Set<String> BLACKLISTED_MEMBERS = Set.of(
+
+  @Value("${hedvig.appleUser.memberId}")
+  private String APPLE_USER_MEMBER_ID;
+
+  private Set<String> BLACKLISTED_MEMBERS = Set.of(
```
palmfjord: should we maybe just put `APPLE_USER_MEMBER_ID` in the blacklisted members set?

src/main/java/com/hedvig/productPricing/service/web/AppleController.java
```diff
@@ -0,0 +1,128 @@
...
+  private String APPLE_USER_LAST_NAME;
+  @Value("${hedvig.appleUser.address.street}")
+  private String APPLE_USER_ADDRESS_STREET;
+  @Value("${hedvig.appleUser.address.city}")
+  private String APPLE_USER_ADDRESS_CITY;
+  @Value("${hedvig.appleUser.address.zipCode}")
+  private String APPLE_USER_ADDRESS_ZIP_CODE;
+
+
+  private final CommandGateway commandBus;
```
palmfjord: ooh final 😬 nice!

src/main/resources/application.yml
```diff
@@ -7,6 +7,15 @@ logging:
...
 
 hedvig:
+  appleUser:
+    address:
+      zipCode: 12345
+      city: California
+      street: Palo Alto
+    firstName: Steve
+    lastName: Jobs
+    personalNumber: 11111
```
palmfjord: I think we should have a valid personnummer here, otherwise we might end up with pretty weird errors no?

src/main/java/com/hedvig/productPricing/service/policies/PolicyEventListener.java
palmfjord: instead of checking it every time? Alternatively i think all those checks should be moved into a util method

PR #162, opened 2019-05-15
Add endpoint to be able to fetch more detailed product information for a member

vonElfvin:


src/main/java/com/hedvig/productPricing/service/service/InsuranceBillingServiceImpl.java
```diff
@@ -266,6 +259,26 @@ public InsuranceBillingDTO getMonthlySubscriptionByMemberId(YearMonth period, St
...
+    public List<InsuranceProductSubscriptionDTO> getMonthlyProductSubscriptionsByMemberId(String memberId, YearMonth period) {
+      val date = LocalDateTime.now().with(period).withDayOfMonth(1).plusMonths(1);
+      List<ProductEntity> productList = productRepository.findBillingByDateByMemberId(date, memberId);
+      return calculateProductSubscription(productList, period);
+    }
+
+    private List<InsuranceProductSubscriptionDTO> calculateProductSubscription(List<ProductEntity> productList, YearMonth period) {
+      Stream<InsuranceBillingInfo> filteredList = productList
+        .stream()
+        .filter(x -> isSignedInsurance(x))
```
palmfjord: no `x`s please :D makes it super-hard to understand what we're checking here

src/main/java/com/hedvig/productPricing/service/service/InsuranceBillingServiceImpl.java
```diff
@@ -266,6 +259,26 @@ public InsuranceBillingDTO getMonthlySubscriptionByMemberId(YearMonth period, St
...
+        .map(InsuranceBillingInfo::new);
+
+      return filteredList.map(info -> new InsuranceProductSubscriptionDTO(
```
palmfjord: just some nitty-gritty but can we move `.map` and `.collect` to newlines? Makes it sliiightly easier to read

src/main/java/com/hedvig/productPricing/service/service/InsuranceBillingServiceImpl.java
```diff
@@ -266,6 +259,26 @@ public InsuranceBillingDTO getMonthlySubscriptionByMemberId(YearMonth period, St
 
     }
 
+    @Override
+    public List<InsuranceProductSubscriptionDTO> getMonthlyProductSubscriptionsByMemberId(String memberId, YearMonth period) {
+      val date = LocalDateTime.now().with(period).withDayOfMonth(1).plusMonths(1);
```
palmfjord: I don't understand this.
a) As i understand it this'll be _next_ month no? Is this expected? It doesn't make sense imo to calculate the price not for the period we're passing but for the period _after_ it, i would expect `getMonthlyProductSubscriptionsByMemberId` to calculate on whatever it receives
b) Why are we using `LocalDateTime.now()`? Why not just `period.atDay(1)` if we want a locadate?

src/main/java/com/hedvig/productPricing/service/web/dto/InsuranceProductSubscriptionDTO.java
```diff
@@ -0,0 +1,22 @@
...
+@Value
+public class InsuranceProductSubscriptionDTO {
+  UUID productId;
+  ProductTypes productType;
+  MonetaryAmount subscription;
+
+  public InsuranceProductSubscriptionDTO(UUID productId, ProductTypes productType, BigDecimal amount) {
+    this.productId = productId;
+    this.productType = productType;
+    this.subscription = Money.of(amount, "SEK");
```
palmfjord: Why do we need this? I feel like we could just pass `MonetaryAmount` directly and then we don't need to write this constructor at all

src/main/java/com/hedvig/productPricing/service/service/InsuranceBillingServiceImpl.java
vonElfvin: Honestly I don't know, I'm just doing as the old methods do it 🙈

palmenhq: That's a terrible excuse 😬 🙈  this is quite critical I believe, as we must charge for the correct month no?

vonElfvin: Given that, as you stated, this is critical, Why would I want to change the date? From my understanding this is thoroughly tested and should this not be tampered with. So for a) no it's most likely the correct date and thus month given this is what we have used all along. And for b) sure that would make it prettier.

vonElfvin: I can double check with Meletis though before resolving this.

PR #163, opened 2019-05-20
WIP Referrals

meletios:


src/main/resources/schema.graphqls
```diff
@@ -0,0 +1,53 @@
...
+    member: Member
+}
+
+type Mutation {
+    ## will be called from the client when 1) redeem manually a code, 2) click the link  --Fails if the code is invalid?--
+    redeemCode(code: String!): Referral
+}
+
+type Member {
+    memberId: String!
```
hedvigoscar: This field may collide with fields upstream I think?

src/main/resources/schema.graphqls
```diff
@@ -0,0 +1,53 @@
+schema {
+    query: Query
+    mutation: Mutation
+}
+
+type Query {
+    member: Member
```
hedvigoscar:
```suggestion
    referralInformation: ReferralInformation
```

src/main/java/com/hedvig/productPricing/service/query/campaign/model/Incentive.kt
```diff
@@ -0,0 +1,17 @@
+package com.hedvig.productPricing.service.query.campaign.model
+
+import java.util.*
+import javax.persistence.Entity
+import javax.persistence.EnumType
+import javax.persistence.Enumerated
+import javax.persistence.Id
+
+@Entity
+data class Incentive(
```
johantjellden: Make incentive and incentive type into classes with inheritance maybe using kotlins sealed class.

PR #166, opened 2019-06-03
update member first and last name in product_pricing when updated in …

AlexSJ1990:
…member_service


src/main/java/com/hedvig/productPricing/service/aggregates/MemberAggregate.java
```diff
@@ -347,6 +325,17 @@ void certificateUpload(CertificateGeneratedCommand cmd) {
     }
   }
 
+  @CommandHandler
+  void NameUpdate(EditMemberNameCommand cmd) {
```
vonElfvin: Usually "NameUpdate" is named "handle" regardless of the command

src/main/java/com/hedvig/productPricing/service/aggregates/MemberAggregate.java
```diff
@@ -347,6 +325,17 @@ void certificateUpload(CertificateGeneratedCommand cmd) {
...
+    if ((cmd.getFirstName() != null
+      && !Objects.equals(this.firstName, cmd.getFirstName()))
+      || (cmd.getLastName() != null
+      && !Objects.equals(this.lastName, cmd.getLastName()))
```
vonElfvin: a help function to make this more understandable would be nice

src/main/java/com/hedvig/productPricing/service/aggregates/MemberAggregate.java
```diff
@@ -468,4 +457,9 @@ public void on(CertificateUploadedEvent e) {
         true;
   }
 
+  @EventSourcingHandler
+  public void on(NameUpdatedEvent e) {
+    this.firstName = e.getFirstName();
+    this.lastName = e.getLastName();
```
vonElfvin: I may be wrong but this would change "lastname" to null if firstName != null and firstName!=this.firstName

Suggestion: Use a ternary operator: 
this.firstName ?: e.getFirstName()
this.lastName ?: e.getLastName()

src/main/java/com/hedvig/productPricing/service/commands/EditMemberNameCommand.java
```diff
@@ -0,0 +1,24 @@
...
+public class EditMemberNameCommand {
+
+  @TargetAggregateIdentifier
+  final private String memberId;
+
+  final private String firstName;
+  final private String lastName;
+
+  public EditMemberNameCommand(
+    String memberId, MemberChangeNameDto body, String token) {
```
vonElfvin: Please be consistent with the word choice, i.e. "edit" or "change" when talking about updating the name of a user. Just stick to one :+1:

src/main/java/com/hedvig/productPricing/service/query/ProductEventListener.java
```diff
@@ -94,6 +80,18 @@ public void on(MemberCreatedEventV2 e) {
...
   }
 
+  @EventHandler
+  public void on(NameUpdatedEvent e) {
+    log.info("Updating a member name: " + e);
+
+    UserEntity userEntity = userRepository.findById(e.getMemberId()).get();
+
+    userEntity.setFirstName(e.getFirstName());
+    userEntity.setLastName(e.getLastName());
```
vonElfvin: Similar to above, ensure this won't set the name to null

src/main/java/com/hedvig/productPricing/service/web/InternalController.java
```diff
@@ -597,4 +597,15 @@ private String generateCertificateURL(ProductEntity product) {
 
     return perilsHashMap;
   }
+
+  @PostMapping("{memberId}/edit")
```
vonElfvin: This name may be too broad for only changing the member's name 🤔

src/main/java/com/hedvig/productPricing/service/web/dto/MemberChangeNameDto.java
```diff
@@ -0,0 +1,11 @@
+package com.hedvig.productPricing.service.web.dto;
+
+import lombok.Value;
+
+@Value
+public class MemberChangeNameDto {
```
vonElfvin: Maybe use "DTO" in caps to be consistent with other Dtos in the service. Also the name could be easier to understand if the order of words is different "ChangeMemberName"

src/main/java/com/hedvig/productPricing/service/commands/EditMemberNameCommand.java
palmenhq: Also why do we even pass the token here? looks like its not used

src/main/java/com/hedvig/productPricing/service/events/NameUpdatedEvent.java
```diff
@@ -0,0 +1,12 @@
+package com.hedvig.productPricing.service.events;
+
+import lombok.AllArgsConstructor;
+import lombok.Value;
+
+@Value
+@AllArgsConstructor
+public class NameUpdatedEvent {
+  private final String memberId;
```
palmenhq: final 😍

src/main/java/com/hedvig/productPricing/service/query/UserEntity.java
```diff
@@ -25,6 +25,13 @@
 @Entity
 @Getter
 public class UserEntity {
+  public void setFirstName(String firstName) {
```
palmenhq: usually setters and getters are put after properties and constructors, so this should prolly go at the end of the class :)

palmenhq:
I feel a bit scared to update the product/policy for legal reasons (especially regarding what the insurance certificate says), have we double-checked this with someone?

vonElfvin:
:+1:

vonElfvin:
wtf

PR #169, opened 2019-06-03
Backfill missing certficates

palmenhq:


src/main/java/com/hedvig/productPricing/service/query/ProductRepository.java
```diff
@@ -103,4 +103,12 @@
       + "order by p.activeFrom DESC"
   )
   List<ProductEntity> findInitiatedProductsByMemberIds(@Param("memberIds") Collection<String> memberIds);
+
+  @Query(
+    "FROM ProductEntity p"
+      + " WHERE p.state != 'QUOTE'"
+      + " AND p.certificateKey IS NOT NULL"
```
meletios: Is not null or is null ?

meletios: Alt Text

src/main/java/com/hedvig/productPricing/service/web/InsuranceCertificateController.java
```diff
@@ -0,0 +1,105 @@
...
+import org.springframework.web.bind.annotation.*;
+import org.springframework.web.multipart.MultipartFile;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Optional;
+import java.util.UUID;
+
+@Slf4j
+@RestController
```
meletios: Shall we mark here, the controller with the uri /i/insurance

src/main/java/com/hedvig/productPricing/service/web/InsuranceCertificateController.java
```diff
@@ -0,0 +1,105 @@
...
+    @Value("${hedvig.productPricing.service.certificatesS3Bucket}") String certificateS3Bucket,
+    final AmazonS3Client s3Client,
+    final InsuranceService insuranceService
+  ) {
+    this.certificateS3Bucket = certificateS3Bucket;
+    this.s3Client = s3Client;
+    this.insuranceService = insuranceService;
+  }
+
+  @PostMapping({"/i/insurance/{memberId}/certificate", "/_/insurance/{memberId}/certificate"})
```
meletios: and then this one can be certificate/{memberId}/upload ? idk

src/main/java/com/hedvig/productPricing/service/web/InsuranceCertificateController.java
```diff
@@ -0,0 +1,105 @@
...
+  public ResponseEntity<?> insuranceCertificate(
+    @PathVariable String memberId, @RequestParam MultipartFile file, @RequestHeader("Authorization") String token
+  ) throws IOException {
+
+    final Optional<ProductEntity> byMemberId = insuranceService.GetCurrentInsurance(memberId);
```
meletios: Shall we rename this to insuranceMaybe or productMaybe?

src/main/java/com/hedvig/productPricing/service/web/InsuranceCertificateController.java
meletios: just saw that this one shouldnt change, since they are coming from the internal controller so discard my comment

src/main/java/com/hedvig/productPricing/service/query/ProductRepository.java
palmenhq: woopsie, nice catch 😬 could've backfired completely

src/main/java/com/hedvig/productPricing/service/web/InsuranceCertificateController.java
palmenhq: Imo no, because it's already kinda claimed by `InternalController` and i don't want them to collide

src/main/java/com/hedvig/productPricing/service/web/InsuranceCertificateController.java
palmenhq: sure, this is just moved blindly from internalcontroller though so maybe its better to not touch it for now?

johantjellden:
Looks good!

PR #183, opened 2019-06-11
Removed old BDX code

johantjellden:


src/main/java/com/hedvig/productPricing/service/web/InternalController.java
```diff
@@ -473,23 +474,33 @@ public ProductSearchResultDTO searchPaged(ProductSearchRequestDTO searchRequest)
...
 
+  /**
+   * Remove this method as soon as the BDX tab is removed from backoffice
+   *
+   * @param year
+   * @param month
+   * @param productType
+   * @return
+   */
+  @Deprecated
```
meletios: Why not remove it from now?

meletios: Since nobody uses this tab is using this tab

johantjellden: There will be an ugly error in hope.

meletios: Correcto! Lets leave it

PR #190, opened 2019-06-14
Fix/bug in get referral campaign member information

FredrikAreschoug:


johantjellden:
I think we should merge this into feature/referral and then merge that into master.

FredrikAreschoug:
Decided to scrap getReferralCampaignMemberInformation

PR #199, opened 2019-06-18
Update campaign member name

FredrikAreschoug:


src/main/java/com/hedvig/productPricing/campaign/query/model/CampaignOwner.kt
```diff
@@ -11,8 +11,8 @@ sealed class CampaignOwner(
 ) {
 
   @Entity
-  class Member(
-    ownerId: UUID,
+  data class Member(
```
johantjellden: Data classes don't work that well with hibernate, we should probably make the attributes non imutable.

FredrikAreschoug: Okay I will fix that! Didn't know about that with hibernate

PR #200, opened 2019-06-18
Feature/roust code generation

johantjellden:


src/test/java/com/hedvig/productPricing/campaign/domain/code/CampaignCodeFactoryTest.kt
```diff
@@ -0,0 +1,62 @@
...
+class CampaignCodeFactoryTest {
+
+  @Mock
+  lateinit var campaignCodeRepository: CampaignCodeRepository
+
+  @Mock
+  lateinit var random: Random
+
+  @Before
+  fun setUp() {
```
meletios: do we need that?

PR #204, opened 2019-06-19
Added type MonetaryAmountV2

johantjellden:


hedvigoscar:
I'd prefer if we deprecate the old MonetaryAmount as we're not using it in production.

This way we do not pollute the schema unnecessarily

PR #209, opened 2019-06-20
apply dis to swi

meletios:


johantjellden:
@meletios should we merge this?

PR #210, opened 2019-06-20
Add use MonetaryAmountV2

johantjellden:


palmenhq:
I believe this will break existing thins in HOPE, we should not remove the `MonetaryAmount` scalar until we've removed all usages

johantjellden:
We are not removing MonetaryAmount from the other schemas. Only from this schema where it is not in use. My reasoning is that during stitching we will not break the other schemas by not having it defined since we are not using it.

palmenhq:
@johantjellden but if we remove the scalar parser we cant parse it anymore?

johantjellden:
Yes but lets not send any MonetaryAmount to this service. When we need to integrate hedgehog cant it send monetaryamountv2 to this service?

palmenhq:
@johantjellden i think it’ll be very confusing but sure. Are we stitching in back-office tho?

PR #219, opened 2019-06-25
Insurance with not only cost

FredrikAreschoug:


src/main/java/com/hedvig/productPricing/service/graphql/scalars/LocalDateScalar.kt
```diff
@@ -0,0 +1,49 @@
+package com.hedvig.productPricing.service.graphql.scalars
```
johantjellden: Do we still need this?

PR #220, opened 2019-06-26
[DO NOT MERGE] Fix policy expiry dates

palmenhq:
Putting this here for my own sanity so I can get an overview of the refactor work, but **this will break a lot of things if it gets merged in its current state**.


palmenhq:
Closing in favor of #221

PR #221, opened 2019-06-26
Fix: Expiry dates & cancellations after product is modified

palmenhq:
Would love to get some more eyes on this, especially this part which i am unsure about: https://github.com/HedvigInsurance/product-pricing/compare/47cd092d817a24a451382a9efb8eafe6eeea3654%5E...a681bc5fe32e54ae6ea5960abf1fa84e3f34dccf

The problems:
a) The expiration date is off [in the manager](https://github.com/HedvigInsurance/product-pricing/commit/7418573c4f9a621a44d395f1d29be9e74b4803a7#diff-0e872ea34875b5adadd7e32452bbcc5aL145) and [in the scheduler](https://github.com/HedvigInsurance/product-pricing/commit/bb2b678b462cc62d123667cc3301ab084fe90347#diff-99630ff06445f54a5ee089512fb74a3bL88) - it should be capped to 1 year ahead but is not right now
b) if we modify a product and then set a cancellation date for the modified product the scheduler will pick it up as if the new product was cancelled which is incorrect. solved with a681bc5 but i'm unsure whether it'll break other things or if i didn't consider something 🤷‍♀️ plz assist 🚑
c) As we all know, we may not change history.
<em>d) As we all also happen know, the world needs more abstract factory classes</em>


src/main/java/com/hedvig/productPricing/service/policies/scheduling/PolicyTermPeriodSchedulerUntil_2019_06_19_Impl.java
```diff
@@ -13,8 +16,13 @@
 import static java.util.Comparator.comparing;
 import static java.util.stream.Collectors.toList;
 
-@Component
-public class PolicyTermPeriodSchedulerImpl implements PolicyTermPeriodScheduler {
+@Deprecated(since = "2019-06-20T00:00:00.0Z")
```
palmenhq: Obviously these dates (and the dates of the class names) must change but idk when we'll get this out

src/main/java/com/hedvig/productPricing/service/policies/scheduling/WhoopsyDaisyAwarePolicyTermPeriodSchedulerFactory.java
```diff
@@ -0,0 +1,19 @@
...
+
+  import java.time.Instant;
+
+@Component
+public class WhoopsyDaisyAwarePolicyTermPeriodSchedulerFactory implements PolicyTermPeriodSchedulerFactory {
+  private static final Instant DATE_OF_ACTIVE_TO_AND_INCEPTION_FIX =  Instant.parse("2019-06-19T00:00:00.0Z");
+  @SuppressWarnings("deprecated")
+  @Override
+  public PolicyTermPeriodScheduler createFor(final Instant timeOfKnowledge) {
+    if (!timeOfKnowledge.isAfter(DATE_OF_ACTIVE_TO_AND_INCEPTION_FIX)) {
```
palmenhq: so technically we'll exclude everything from midnight `2019-06-19`? This should probably be `timeOfKnowledge.isBefore(LocalDate.parse("2019-06-20T00:00:00.0Z"))`

palmenhq:
@HedvigInsurance/platform plz review again, would like to get this out asap :)

palmenhq:
@vonElfvin i think 10+11 should be fine no? Since 10 (today) is the last day and 11 (tomorrow) is the first day of the change

PR #225, opened 2019-06-28
preparing for the referrals release, add insurnaceCost type

meletios:


src/main/java/com/hedvig/productPricing/service/web/dto/InsuranceCost.java
```diff
@@ -0,0 +1,22 @@
...
+public class InsuranceCost {
+
+  private static final String SEK = "SEK";
+
+  private MonetaryAmountV2 monthlyGross;
+  private MonetaryAmountV2 monthlyDiscount;
+  private MonetaryAmountV2 monthlyNet;
+
+  InsuranceCost(BigDecimal currentTotalPrice) {
+    this.monthlyGross = MonetaryAmountV2.Companion.of(currentTotalPrice, SEK);
```
palmenhq: This is not super-beautiful, maybe we could mark these as `@JvmStatic` so we can use them like "regular" static methods from java?

meletios: Agree! i will not change it now though

meletios: will fix it with the release of referrals

PR #232, opened 2019-07-02
deciding redeeemed campaign status according to redemmer and camapign…

meletios:
… owner


src/main/java/com/hedvig/productPricing/campaign/domain/model/RedeemedCampaign.kt
```diff
@@ -44,24 +55,11 @@ open class RedeemedCampaign(
...
-
-      val discountCheckerOk = when (discountChecker) {
-        this.campaign.campaignOwner -> this.getStatus() == RedeemedCampaignStatus.ACTIVE
-        this.redeemedBy -> true
-        else -> return false
-      }
-
-      val campaignOwnerSigned = (this.campaign.campaignOwner as Member).getMemberStatus() == MemberStatus.SIGNED
-      return discountCheckerOk && campaignOwnerSigned
+  fun shouldProvideDiscount(): Boolean {
```
johantjellden: Does this still work for both referrer and referent? Dont they calculate the discounts differently when a referent is not signed?

PR #233, opened 2019-07-02
[WIP] remove medlems email notification when we don't know members cur…

AlexSJ1990:


src/main/java/com/hedvig/productPricing/service/service/InsuranceNotificationServiceImpl.java
```diff
@@ -44,31 +44,14 @@ public void sendInsuranceSignedNotification(
     if (!insuredAtOtherCompany) {
       notificationService.insuranceSigned(memberId, false, null);
     } else {
-      if (currentInsurer == null || currentInsurer.equals(InsuranceCompaniesSE.OTHER)) {
-        notifyUnknownInsurer(memberId);
+      if (currentInsurer == null) {
```
AlexSJ1990: how can I make this better and get rid of and if else within an else - as cannot use early return as returning void?

palmenhq: Yes that sounds very nice! Nested `if`s = blä 😅 i think early returns would solve it

PR #244, opened 2019-07-10
use already existing policy, if it was previously modified

meletios:


src/main/java/com/hedvig/productPricing/service/policies/PolicyEventListener.java
```diff
@@ -51,20 +51,27 @@ public void on(final ContractSignedEvent e, final @Timestamp Instant timestamp)
     if (e.getActiveFrom() == null) {
       log.info("Received contract signed event without activation date, skipping [memberId={}]", e.getMemberId());
       return;
+    } else {
+      final LocalDate normalizedActiveFrom = LocalDate.ofInstant(e.getActiveFrom().toInstant(ZoneOffset.UTC), STOCKHOLM);
```
palmenhq: What's wrong with `e.getActiveFrom().toLocalDate()` ? 😬 are we sure the TZ is off here?

PR #261, opened 2019-07-25
Renew products + certificates

palmenhq:
This is a small step for man, a big step for renewals.


src/main/java/com/hedvig/productPricing/service/policiesv2/PolicyV2EventListener.kt
```diff
@@ -22,18 +27,79 @@ class PolicyV2EventListener @Autowired constructor(
...
-  fun on(event: ProductModifiedEvent, @Timestamp timeOfKnowledge: Instant) = handleEvent(event.memberId, timeOfKnowledge, ProductModifiedEvent::class.simpleName!!)
+  fun on(
+    event: ProductModifiedEvent,
+    @Timestamp timeOfKnowledge: Instant,
+    replayStatus: ReplayStatus
+  ) {
+    val result = handleEvent(event.memberId, timeOfKnowledge, ProductModifiedEvent::class.simpleName!!)
+
+    if (replayStatus.isReplay) {
+      return
```
AlexSJ1990: Is this just so that if we replay events at some point it won't re-trigger the certificates to be generated?

palmenhq: Yes exactly 👍

src/main/java/com/hedvig/productPricing/service/policiesv2/scheduling/PolicyScheduler.kt
```diff
@@ -142,6 +149,30 @@ class PolicyScheduler(
...
+
+    val newPolicy = createPolicy(
+      product = productToRenew,
+      memberId = state.memberId,
+      activeFrom = event.renewalDate,
+      timestamp = timestamp
+    )
+    newPolicy.previousPolicy = originalPolicy
+    state.policies.add(newPolicy)
+    // TODO do we need to somehow terminate the previous policy?
```
AlexSJ1990: perhaps we could have an expired state

src/main/java/com/hedvig/productPricing/service/query/ProductEventListener.java
```diff
@@ -210,6 +209,44 @@ public void on(ModifiedProductCreatedEvent e) {
...
+      .collect(Collectors.toList());
+
+    final ProductEntity originalProduct = repository.findById(event.getOriginalProductId())
+      .orElseThrow(() -> new IllegalStateException("Tried to find non-existing original product " + event.getOriginalProductId() + " while renewing it"));
+
+    renewedProduct.activeFrom = event.getRenewalDate().atStartOfDay();
+    final boolean shouldTransferTerminationDateToRenewal = originalProduct.activeFrom != null
+      && !originalProduct.activeFrom.isBefore(event.getRenewalDate().plusDays(1).atStartOfDay());
+    if (shouldTransferTerminationDateToRenewal) {
+      renewedProduct.activeTo = originalProduct.activeTo;
```
AlexSJ1990: would this situation only occur if we have been made aware that a member will terminate their policy for example 3 months after the renewal date?

palmenhq: Yes exactly

src/main/java/com/hedvig/productPricing/service/policiesv2/scheduling/PolicyScheduler.kt
```diff
@@ -142,6 +149,30 @@ class PolicyScheduler(
     return state
   }
 
+  private fun plan(state: MemberStateSnapshot, event: ProductRenewedEvent, timestamp: Instant): MemberStateSnapshot {
+    val originalProduct = findProductWithId(event.originalProductId)
+    val productToRenew = findProductWithId(event.newProductId)
```
vonElfvin: I may be wrong but in my head this is the the "newProduct" and the "originalProduct" is the "productToRenew"

src/main/java/com/hedvig/productPricing/service/policiesv2/scheduling/PolicyScheduler.kt
vonElfvin: If the renewal process is done correctly the expiry date should be today and thus it is terminated by "by design".

src/main/java/com/hedvig/productPricing/service/query/ProductEventListener.java
```diff
@@ -210,6 +209,44 @@ public void on(ModifiedProductCreatedEvent e) {
...
+    renewedProduct.cancellationEmailSentAt = null; // Do we still need this?
+    renewedProduct.goodToHaveItems = event.getSafetyIncreasers().stream()
+      .map(SafetyIncreaser::getName)
+      .collect(Collectors.toList());
+
+    final ProductEntity originalProduct = repository.findById(event.getOriginalProductId())
+      .orElseThrow(() -> new IllegalStateException("Tried to find non-existing original product " + event.getOriginalProductId() + " while renewing it"));
+
+    renewedProduct.activeFrom = event.getRenewalDate().atStartOfDay();
+    final boolean shouldTransferTerminationDateToRenewal = originalProduct.activeFrom != null
```
vonElfvin: should it be "originalProduct.activeTo != null"?

src/main/java/com/hedvig/productPricing/service/web/InternalProductController.java
```diff
@@ -0,0 +1,70 @@
...
+  @Autowired
+  public InternalProductController(
+    final CommandGateway commandGateway,
+    final ProductRepository productRepository
+  ) {
+    this.commandGateway = commandGateway;
+    this.productRepository = productRepository;
+  }
+
+  @PostMapping("/_/products/{productIdToRenew}/renew")
```
vonElfvin: This solution is not the final one I presume.

src/main/java/com/hedvig/productPricing/service/policiesv2/scheduling/PolicyScheduler.kt
palmenhq: hehe this naming is super weird i agree, will refactor

src/main/java/com/hedvig/productPricing/service/query/ProductEventListener.java
palmenhq: yes ofc 😂

src/main/java/com/hedvig/productPricing/service/policiesv2/scheduling/PolicyScheduler.kt
vonElfvin: I think we concluded that the only way to add a policy to a chain is by renewal. So if there is a new policy in the chain it has been properly renewed. I.e. if there is a policy who's previous policy is the current policy then this policy has been renewed.

vonElfvin: I may be missing something though

PR #263, opened 2019-07-30
override terminationDate if we modify the product

meletios:


src/main/java/com/hedvig/productPricing/campaign/domain/model/Member.kt
```diff
@@ -148,6 +148,11 @@ open class Member(
     this.getCurrentRedeemedCampaign()?.setMemberTerminationDate(terminationDate)
   }
 
+  fun clearTermiantionDate() {
```
hedvigoscar:
```suggestion
  fun clearTerminationDate() {
```

meletios: thanks 😄

PR #267, opened 2019-07-31
Feature/renewal notifications

vonElfvin:
Ignore all files in campaign, it's not me who have touched them it's just git acting up


src/main/java/com/hedvig/productPricing/service/aggregates/MemberAggregate.java
```diff
@@ -409,10 +442,6 @@ public void on(ModifiedProductCreatedEvent e) {
 
   @EventSourcingHandler
   public void on(final ProductRenewedEvent event) {
-    if (this.products == null) {
```
AlexSJ1990: why do we delete this?

src/main/java/com/hedvig/productPricing/service/web/InternalRenewalController.kt
```diff
@@ -0,0 +1,60 @@
...
+@RequestMapping("/_/renewals")
+class InternalRenewalController @Autowired constructor(
+  private val commandGateway: CommandGateway,
+  private val renewalFacade: RenewalQueueFacade
+) {
+
+  @PostMapping("/queue-renewals")
+  fun queueRenewals(): ResponseEntity<Void> {
+    val policiesToPotentiallyRenew = renewalFacade.findPoliciesToRenew(
+      renewOnOrAfter = LocalDate.of(2019, 2, 1),
```
AlexSJ1990: why do we hard-code 2019-02-01?

src/main/java/com/hedvig/productPricing/service/aggregates/MemberAggregate.java
vonElfvin: It's impossible to renew a product that doesn't exist

PR #287, opened 2019-08-19
[PoC] Update the logic when scheduling policies to care about chains of products instead

vonElfvin:
This way of scheduling policies ensures no person can have parallel policies when they shouldn't. Can be worth considering using this way of scheduling instead. It passes all PolicySchedulerTests.


src/main/java/com/hedvig/productPricing/service/policiesv2/Product.kt
```diff
@@ -11,5 +12,8 @@ data class Product(
   val productType: ProductTypes,
   var activeFrom: LocalDate? = null,
   var activeTo: LocalDate? = null,
-  var modifiedByProduct: Product? = null
+  var wasRenewed: Boolean = false,
+  var modifiedByProduct: Product? = null,
+  var wasSigned: Boolean = false,
+  var lastUpdatedAt: Instant = Instant.now()
```
palmenhq: I _really_ don't like these types of fields as it makes the data model lossy. Can we implement some kind of history table or something instead? Just to not lose data

vonElfvin: Sure

PR #289, opened 2019-08-20
Fix generation of (draft) certificates of legacy members

vonElfvin:


src/main/java/com/hedvig/productPricing/service/aggregates/MemberAggregate.java
```diff
@@ -177,10 +177,10 @@ public void handle(final RenewProductCommand cmd) {
         cmd.getRenewalDate(),
         productToRenew.currentInsurer,
         false,
-        productToRenew.address,
-        productToRenew.livingSpace,
+        productToRenew.address != null ? productToRenew.address : this.address,
```
palmenhq: Just a thought, should we add some kind of flag to the event that we copied the address from the member - just in case something is fishy in the future.

src/test/java/com/hedvig/productPricing/service/renewals/RenewalQueueManagerTest.kt
```diff
@@ -107,6 +107,70 @@ class RenewalQueueManagerTest {
...
+    }
+    every { insuranceCertificateGenerator.generate(any()) } answers {
+      arg<InsuranceCertificate>(0).apply {
+        setCertificateBucket("bucket")
+        setCertificateKey("key")
+      }
+      mockk()
+    }
+    val result = renewalQueueManager.generateDraftRenewalCertificateUnlessAlreadyQueued("123", productId, expiry.plusDays(1))
+    assertThat(result).isNotNull()
```
johantjellden: You dont need this assertion

PR #290, opened 2019-08-20
Feature/campaign voucher billing

johantjellden:
WIP for now. I have taken a lot of inspiration from [Working Effectively with Unit Tests](https://www.amazon.com/Working-Effectively-Unit-Tests-Fields-ebook/dp/B00QS2HXUO)


src/main/java/com/hedvig/productPricing/campaign/domain/model/Incentive.kt
```diff
@@ -0,0 +1,60 @@
...
+    val activeDaysInPeriod =
+      ProRataCalculator.calculateNumberOfDaysInPeriod(
+        DateUtil.toLocalDate(activatedAt),
+        if (activeTo != null)
+          DateUtil.toLocalDate(activeTo)
+        else
+          monthToCheck.atEndOfMonth(),
+        monthToCheck
+      )
+    return discountPerDay.multiply(BigDecimal.valueOf(activeDaysInPeriod))//.setScale(10, RoundingMode.HALF_EVEN)
```
palmenhq: 🙈

src/main/java/com/hedvig/productPricing/campaign/domain/model/Member.kt
```diff
@@ -156,4 +159,73 @@ open class Member(
   fun hasRedeemedCampaignFrom(redeemer: Member): Boolean {
     return this.getCurrentRedeemedCampaign()?.campaign?.campaignOwner == redeemer
   }
+
+  fun getReferralKickback(monthToCheck: YearMonth): BigDecimal {
+    return referralCampaign!!.kickback(monthToCheck)
+  }
+
+  fun getDiscountFromReferralCampaign(month: YearMonth): BigDecimal? {
```
palmenhq: why is this nullable? It'll just return `BigDecimal.ZERO` if there's no discounts, no?

src/main/java/com/hedvig/productPricing/campaign/domain/model/ReferralCampaign.kt
```diff
@@ -20,4 +21,22 @@ open class ReferralCampaign(campaignId: UUID,
...
+
+  fun kickback(monthToCheck: YearMonth): BigDecimal {
+    val validRedemptions = redemptions.filter { it.incentive.type == IncentiveType.COST_DEDUCTION
+      && it.shouldProvideDiscountWithinMonth(
+      discountChecker = this.campaignOwner as Member,
+      monthOfCheck = monthToCheck,
+      billingDiscount = true
+    ) }
+
+    return validRedemptions.map {
```
palmenhq: this is very nitty gritty but imo it's more readable to chain the streaming api on separate lines, like so:

```kt
validRedemptions
  .map { ... }
  .foldRight(...) {...}
```

src/main/java/com/hedvig/productPricing/campaign/service/CampaignBillingServiceImpl.kt
```diff
@@ -70,127 +68,50 @@ class CampaignBillingServiceImpl(
...
-    val remainingToZero = monthlyCost.subtract(totalReferralDiscount)
-
-    if (remainingToZero <= BigDecimal.ZERO)
-      return null
-
-    return VoucherDiscount(remainingToZero, totalActiveVoucherDiscount.first().campaign.code)
+  ): Pair<BigDecimal, String>? {
+
+    val discountsThatGiveFreeMonth = member.getFreeMonthCampaign(month, monthlyCost)
+    return if (discountsThatGiveFreeMonth != null) {
```
palmenhq: to make this more readable i think we could do something like

```kt
return dicountsThatGiveFreeMonth
  ?.filter { monthlyCost.minus(totalReferralDiscount) > BigDecimal.ZERO }
  ?.let {
    Pair(...stuff...)
  }
```
(may require some more thinking but i think it would be more clean)

palmenhq: (i don't like ifs 😬)

src/main/java/com/hedvig/productPricing/campaign/domain/CampaignOwnerRepository.kt
```diff
@@ -17,4 +17,8 @@ interface CampaignOwnerRepository : JpaRepository<CampaignOwner, String> {
   @Query("select c from CampaignOwner c where c.memberId = ?1")
   @EntityGraph(attributePaths = ["redeemedCampaigns", "redeemedCampaigns.campaign.campaignOwner"])
   fun findByMemberIdLoadRedeemedCampaignsAndTheirOwners(memberId: String): Member?
+
+  @Query("select c from CampaignOwner c where c.memberId = ?1")
+  @EntityGraph(attributePaths = ["redeemedCampaigns.campaign.campaignOwner", "referralCampaign.redemptions"])
+  fun findByMemberIdLoadRedeemedCampaignsWithOwnersAndReferralCampaignRedemptions(memberId: String): Member?
```
palmenhq: death my method name length 😂 better explicit than not but its crazy long

src/main/java/com/hedvig/productPricing/campaign/domain/model/RedeemedCampaign.kt
```diff
@@ -89,52 +71,36 @@ open class RedeemedCampaign(
...
-        billingDiscount
-      )
+  private fun shouldProvideDiscount(discountChecker: Member, timeOfCheck: Instant, endOfMonth: Instant, billingDiscount: Boolean): Boolean {
+    return when (this.campaign) {
+      is ReferralCampaign -> when (discountChecker) {
+        this.campaign.campaignOwner -> this.getStatus(timeOfCheck, endOfMonth) == RedeemedCampaignStatus.ACTIVE
+        this.redeemedBy ->
+          this.getStatus(timeOfCheck, endOfMonth) == RedeemedCampaignStatus.ACTIVE
+            || (this.getStatus(timeOfCheck, endOfMonth) == RedeemedCampaignStatus.OFFER && !billingDiscount)
+        else -> throw IllegalArgumentException("DiscountChecker is neither the campaignOwner or the redeemer")
```
palmenhq: Maybe we can add something about actual value and expected values also? It'll be hard to track down the cause otherwise if this happens

PR #296, opened 2019-09-02
Add segment listener for when a member's product is cancelled

vonElfvin:
We are now incorrectly assuming that a member who has an active from is an active member in customer.io. Extending the segment with "not existing" cancellation date/a cancellation date in the future would make the segment more correct. The current use case is to remind people to connect DD and not send that email to terminated members.


src/main/java/com/hedvig/productPricing/service/service/segmentExporter/ProductCancelledTrait.kt
```diff
@@ -0,0 +1,11 @@
+package com.hedvig.productPricing.service.service.segmentExporter
+
+import com.fasterxml.jackson.annotation.JsonProperty
+
+class ProductCancelledTrait(
+  @JsonProperty("cancellation_date")
```
palmenhq: could we use `@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy::class)` instead on the class instead?

meletios:
check the tests though

src/main/java/com/hedvig/productPricing/service/service/segmentExporter/ProductCancelledTrait.kt
vonElfvin: Fixed!

PR #299, opened 2019-09-03
Fix a bunch of issues with dates

johantjellden:


meletios:
It seems that there is a conflicting file from master. can you check?

PR #301, opened 2019-09-05
Add the hack for extracting reports data from renewals

palmenhq:
it breaks my heart, but we need to check this in


src/main/java/com/hedvig/productPricing/service/service/reports/RenewalsMiReportHack.java
```diff
@@ -0,0 +1,174 @@
...
+    this.entityManagerFactory = requireNonNull(entityManagerFactory);
+    this.productRepository = requireNonNull(productRepository);
+  }
+
+  public Map<YearMonth, PartialMiReportHackData> generateHackedData() {
+    final List<ProductRenewedEvent> productRenewedEvents = new ArrayList<>();
+
+    try (final Session session = entityManagerFactory.unwrap(SessionFactory.class).openSession()) {
+      session.doWork(connection -> {
+        final ResultSet result = connection.prepareStatement(
```
meletios: Should we read all the events up to the end of the current reporting month?

palmenhq: Sure, i mean it won't matter i think since we always perform the renewals the same day but better safe than sorry

PR #314, opened 2019-09-16
Gracefully fail when a member's policy does not have policy term periods

vonElfvin:


palmenhq:
although #313 is merged now so that should fix it i guess

PR #315, opened 2019-09-17
Localize perils

FredrikAreschoug:


src/main/java/com/hedvig/productPricing/service/web/dto/InsuranceDTO.java
```diff
@@ -40,15 +41,15 @@
...
   }
 
-  public InsuranceDTO(ProductEntity p, List<PerilEntity> perils, UserEntity ue, InsuranceCost cost) {
-    this(Clock.systemDefaultZone(), p, perils, ue, cost);
+  public InsuranceDTO(ProductEntity p, List<PerilEntity> perils, UserEntity ue, InsuranceCost cost, Locale locale, LocalizationService localizationService) {
+    this(Clock.systemDefaultZone(), p, perils, ue, cost, locale, localizationService);
   }
 
-  public InsuranceDTO(Clock clock, ProductEntity p, List<PerilEntity> perils, UserEntity ue, InsuranceCost cost) {
+  public InsuranceDTO(Clock clock, ProductEntity p, List<PerilEntity> perils, UserEntity ue, InsuranceCost cost, Locale locale, LocalizationService localizationService) {
```
palmenhq: Maybe we could move `localizationService` into a class prop (and make it `@Transient`/`@JsonIgnore`) and use that in `getLocalization` instead of passing it around so much?

src/main/java/com/hedvig/productPricing/service/web/dto/InsuranceDTO.java
```diff
@@ -38,6 +41,11 @@
   public Float livingSpace;
   public Renewal renewal;
 
+  @JsonIgnore
```
palmenhq: This is awesome but i think we should keep `@Transient` also just so we don't accidentally serialize something if we would use a java serializer :D

PR #327, opened 2019-09-24
Add support for manually redeeming campaigns for members

johantjellden:


src/test/java/com/hedvig/productPricing/campaign/domain/DomainServiceTest.kt
```diff
@@ -131,6 +132,46 @@ class DomainServiceTest {
     assertThat(memberUnderTest.getCurrentRedeemedCampaign()?.campaign).isEqualTo(campaign)
   }
 
+  @Test
+  @Ignore
```
meletios: Shall we remove the Ignore annotation here?

johantjellden: I need to test i dev that this works, so i will keep the test as ignored so that I don't need to dig in git if we need the create campaignOwner functionality.

src/main/java/com/hedvig/productPricing/campaign/domain/model/Campaign.kt
```diff
@@ -36,7 +36,7 @@ open class Campaign(
   )
   open var redemptions: MutableSet<RedeemedCampaign> = mutableSetOf()
 
-  open fun validateRedemption(redeemer: Member): Boolean {
+  open fun validateRedemption(redeemer: Member, dateToCheck: Instant): Boolean {
```
palmenhq: could we not just make this an abstract function instead or?

src/main/java/com/hedvig/productPricing/campaign/domain/model/Member.kt
```diff
@@ -103,15 +112,23 @@ open class Member(
...
+          this.insuranceActivatedAt?.let { insuranceActivatedAt -> redeemedCampaign.activate(insuranceActivatedAt) }
+        }
       }
     }
 
+    campaign.redemptions.add(redeemedCampaign)
+
+
+
+
```
palmenhq: why so many line breaks?

src/main/java/com/hedvig/productPricing/campaign/web/MemberCampaignController.kt
```diff
@@ -45,4 +46,15 @@ class MemberCampaignController(
     }
     return accepted().build()
   }
+
+  @PostMapping(value = ["/addCampaign"])
```
palmenhq: `@PostMapping("/addCampaign")` should be sufficient no?

src/main/java/com/hedvig/productPricing/campaign/domain/model/Member.kt
```diff
@@ -68,7 +72,12 @@ open class Member(
   fun getCurrentRedeemedCampaign() =
     this.redeemedCampaigns.find { redeemedCampaign -> redeemedCampaign.getStatus() != RedeemedCampaignStatus.UNREDEEMED }
 
-  fun redeemCampaign(campaign: Campaign): RedeemedCampaign {
+  fun redeemCampaign(campaign: Campaign, activationDate: LocalDate? = null): RedeemedCampaign {
+    if (activationDate != null) {
+      val firstDayOfMonth = LocalDate.now().withDayOfMonth(1)
+      require(activationDate.isEqual(firstDayOfMonth) || activationDate.isAfter(firstDayOfMonth))
```
meletios:
```suggestion
      require(!activationDate.isBefore(firstDayOfMonth))
```

src/test/java/com/hedvig/productPricing/campaign/domain/DomainServiceTest.kt
palmenhq: Maybe we could give the reason? Just so we remember this. `@Ignore("need to test i dev that this works because blablabla")`

src/test/java/com/hedvig/productPricing/campaign/domain/model/MemberTest.kt
```diff
@@ -402,6 +406,48 @@ class MemberTest {
...
+    val cut = a.MemberBuilder(signedAt = FIRST_MAY_2019, insuranceActivatedAt = FIRST_MAY_2019).build()
+
+    val partner = a.PartnerBuilder().w(a.VoucherCampaignBuilder()).build()
+
+    val activationDate = LocalDate.now().withDayOfMonth(1)
+    val result = cut.redeemCampaign(partner.voucherCampaigns.first(), activationDate = activationDate)
+
+    assertThat(result.campaign).isEqualTo(partner.voucherCampaigns.first())
+    assertThat(result.activatedAt).isEqualTo(activationDate.atStartOfDay(ZoneId.of("Europe/Stockholm")).toInstant())
+    assertThat(result.redeemedAt).isCloseTo(Instant.now(), within(2, ChronoUnit.SECONDS))
```
palmenhq: ): this is very clever but feels bad, could we give ourselves control over time instead?

src/main/java/com/hedvig/productPricing/campaign/domain/model/Campaign.kt
johantjellden: Yes it is a good idea, but requires a bit to much changes for now.

src/main/java/com/hedvig/productPricing/campaign/web/MemberCampaignController.kt
meletios: I guess also the name can be `redeemCampaign` since we are not adding a campaign

src/main/java/com/hedvig/productPricing/campaign/domain/model/Member.kt
johantjellden: Sure, i like the explicitness of the other one more, but i will change.

PR #341, opened 2019-10-14
Fix bug with sending the translated insurance terms of conditions

vonElfvin:
🤦‍♂


src/main/java/com/hedvig/productPricing/service/service/LocalizationService.kt
```diff
@@ -77,7 +77,7 @@ class LocalizationService @Autowired constructor(val localizationClient: Localiz
   }
 
   private enum class Language(val localeLanguage: String, val graphCMSLanguage: String, val fileIdentifier: String) {
-    ENGLISH("en", "en_SE", ""),
-    SWEDISH("sv", "sv_SE", " (English)")
+    ENGLISH("en", "en_SE", " (English)"),
```
FredrikAreschoug: why is there an space?

src/main/java/com/hedvig/productPricing/service/service/PerilsInjector.java
```diff
@@ -303,7 +303,7 @@ public void generatePerils() {
 
     PerilDTO house_house_water= new PerilDTO(
       "HOUSE.HOUSE.WATER",
-      "Vattenläcka",
+      "Vatten-\nläcka",
```
FredrikAreschoug: we will need to change this in graphcms as well

src/main/java/com/hedvig/productPricing/service/service/LocalizationService.kt
vonElfvin: Yes

PR #317, opened 2019-09-17
Contract - Agreements

meletios:

PR #344, opened 2019-10-15
Contract/agreements

palmenhq:
replaces #317 with a prettier branch name


src/main/java/com/hedvig/contract/shared/aggregates/BaseContract.kt
```diff
@@ -0,0 +1,453 @@
...
+    agreements[event.agreementId] = agreements[event.agreementId].copy_(to = event.newTo)
+  }
+
+  @CommandHandler
+  fun handle(command: UpdateInsuranceCertificate) {
+    apply(
+      InsuranceCertificateUpdated(
+        command.contractId,
+        command.agreementId,
+        command.insuranceCertificatePdf
```
vonElfvin: Are we storing the whole pdf in the event?

src/main/java/com/hedvig/contract/shared/aggregates/BaseContract.kt
```diff
@@ -0,0 +1,453 @@
...
+    )
+  }
+
+  @EventSourcingHandler
+  fun on(event: InsuranceCertificateUpdated) {
+    agreements[event.agreementId] = agreements[event.agreementId].copy_(certificate = event.insuranceCertificatePdf)
+  }
+
+  @CommandHandler
+  fun handle(command: QueueAgreementForRenewal) {
```
vonElfvin: Given how the timeline looks to work (from what I can understand that is) why does the renewal date come externally? Shouldnt it be renewed at the date the timeline of agreements end, or am I missing something?

vonElfvin: The logic of a renewal should IMO be controlled by the aggregate itself since it knows which agreement that is up for renewal.

src/main/java/com/hedvig/contract/shared/aggregates/BaseContract.kt
```diff
@@ -0,0 +1,453 @@
...
+      )
+    }
+  }
+
+  @EventSourcingHandler
+  fun on(event: AgreementQueuedForRenewal) {
+    agreements[event.agreementId] = agreements[event.agreementId]
+      .copy_(
+        renewAt = event.renewalDate,
+        renewalDraftCertificate = event.draftCertificatePdf
```
vonElfvin: Storing of information about the renewal should probably not be attached to renewal. E.g. if the agreement is replaced the renewal must be changed to a new agreement and thus storing it on an agreement seems weird to me.

src/main/java/com/hedvig/contract/shared/aggregates/BaseContract.kt
```diff
@@ -0,0 +1,453 @@
...
+    val hasExtendedTo = agreements[command.agreementIdToRenew].extendedTo != null
+    val hasSuccessor = indexOfRenewedAgreement < agreements.size - 1
+    val renewedAgreement =
+      when {
+        hasExtendedTo -> command.newAgreement.copy_<TAgreement>(
+          to = agreements[command.agreementIdToRenew].extendedTo ?: command.newAgreement.to
+        )
+        hasSuccessor -> command.newAgreement.copy_<TAgreement>(
+          to = agreements[indexOfRenewedAgreement + 1].partBeginning.minusDays(1)
+        )
```
vonElfvin: Do you update the active to of previous agreement?

src/main/java/com/hedvig/contract/valueObjects/CoInsured.kt
```diff
@@ -0,0 +1,5 @@
+package com.hedvig.contract.valueObjects
+
+data class CoInsured(
+  val memberId: String? = null
```
vonElfvin: Add ssn here?

vonElfvin: Or rather, how to handle when the CoInsured isn't a member

src/main/java/com/hedvig/contract/valueObjects/InsuranceAddOn.kt
```diff
@@ -0,0 +1,26 @@
...
+) : InsuranceAddOn
+
+
+enum class ObjectType {
+  Watch,
+  Jewellery,
+  Bike,
+  Painting,
+  Other
+}
```
vonElfvin: Separate files maybe

src/main/java/com/hedvig/contract/shared/commands/ReviveContract.kt
```diff
@@ -0,0 +1,11 @@
+package com.hedvig.contract.shared.commands
+
+import com.hedvig.contract.stereotypes.AgreementStereotype
+import org.axonframework.commandhandling.TargetAggregateIdentifier
+import java.util.*
+
+class ReviveContract<TAgreement : AgreementStereotype>(
+  @TargetAggregateIdentifier
+  val contractId: UUID,
+  val nextAgreement: TAgreement
```
vonElfvin: Maybe it is easier to have revival separated from the addition of a new agreement. A renewal could be an act to revive it completely and then you can "go from there" instead of doing it all at once. Thoughts?

src/main/java/com/hedvig/contract/apartments/aggregates/ApartmentContract.kt
```diff
@@ -0,0 +1,28 @@
...
+import com.hedvig.contract.shared.aggregates.BaseContract
+import org.axonframework.commandhandling.CommandHandler
+import org.axonframework.commandhandling.model.AggregateLifecycle.apply
+import org.axonframework.spring.stereotype.Aggregate
+
+@Aggregate
+class ApartmentContract() : BaseContract<ApartmentAgreement>() {
+  @CommandHandler
+  constructor(cmd: CreateContractWithAgreement<ApartmentAgreement>) : this() {
+    if (cmd.agreement.from == null) {
```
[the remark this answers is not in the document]
palmenhq: I would say the reverse, currently we only have switchers where the dates aren’t set, but the definition of a pending agreement is really that we don’t know the dates yet. Idk what the reason would be to lock ourselves into only being able to handle switchers

src/main/java/com/hedvig/contract/shared/aggregates/BaseContract.kt
```diff
@@ -0,0 +1,453 @@
...
+        agreementToActivate.to!!
+      )
+    )
+  }
+
+  @EventSourcingHandler
+  fun on(e: ContractWithActiveAgreementCreated<TAgreement>) {
+    this.id = e.contractId
+    this.holderMemberId = e.holderMemberId
+    this.switchedFrom = e.switchedFrom
```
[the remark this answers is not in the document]
palmenhq: Again, if we for some reason would want to set dates on a switcher, I don’t see why we should block ourselves from doing it

src/main/java/com/hedvig/contract/shared/aggregates/BaseContract.kt
```diff
@@ -0,0 +1,453 @@
...
+
+    apply(
+      AgreementAddedWithPredecessorAdjusted(
+        command.contractId,
+        command.agreement,
+        newPredecessor.id,
+        newPredecessor.partEnd
+      )
+    )
+  }
```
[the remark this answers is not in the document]
palmenhq: In a lot of cases we just want to add an agreement and adjust the previous one so the added agreement “fits” so we don’t get these manual errors with date adjustments for example. It’s hard. Or like what’s the question?

src/main/java/com/hedvig/contract/shared/aggregates/BaseContract.kt
```diff
@@ -0,0 +1,453 @@
...
+  fun on(e: AgreementTerminated) {
+    agreements[e.agreementId] = agreements[e.agreementId].copy_(
+      to = e.termination,
+      isTerminated = true
+    )
+  }
+
+  @EventSourcingHandler
+  fun on(e: AgreementTerminatedAfterExpiry) {
+    agreements[e.agreementId] = agreements[e.agreementId].copy_(extendedTo = e.termination)
```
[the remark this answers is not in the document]
palmenhq: Yes exactly

src/main/java/com/hedvig/contract/shared/aggregates/BaseContract.kt
```diff
@@ -0,0 +1,453 @@
...
+  }
+
+  @EventSourcingHandler
+  fun on(event: ContractRevived<TAgreement>) {
+    agreements += event.nextAgreement
+  }
+
+  @CommandHandler
+  fun handle(cmd: ChangeFromDateOnAgreement) {
+    if (pendingAgreement?.id == cmd.agreementId) {
```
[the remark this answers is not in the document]
palmenhq: No, IMO we should have a sanity check so that we know we’re explicitly target the pending agreement. Even though they _should_ never co-exist weird things happen in a system every now and then haha

src/main/java/com/hedvig/contract/shared/aggregates/BaseContract.kt
```diff
@@ -0,0 +1,453 @@
...
+      val predecessor = agreements[agreementIndex - 1]
+      ensureValid(predecessor.copy_(to = cmd.newFrom.minusDays(1)))
+      apply(
+        ContinuousFromDateOnAgreementsChanged(
+          cmd.contractId,
+          cmd.agreementId,
+          cmd.newFrom,
+          agreements[agreementIndex - 1].id,
+          cmd.newFrom.minusDays(1)
+        )
```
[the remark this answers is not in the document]
palmenhq: Yes, but that makes it even harder as we then need to dispatch them in the correct order to not end up with invalid states, where also the order is different depending on the state. I don’t think we should be afraid of creating different events for different business cases

src/main/java/com/hedvig/contract/shared/aggregates/BaseContract.kt
palmenhq: No, just the reference

src/main/java/com/hedvig/contract/shared/aggregates/BaseContract.kt
palmenhq: No, but if there’s an agreement that “takes over” the renewed agreement (if we scheduled a change for example 1 month after expiry) then we need to renew the agreement for 1 month until the new one starts

src/main/java/com/hedvig/contract/shared/aggregates/BaseContract.kt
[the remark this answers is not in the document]
meletios: The majority of the cases are non-switchers, yes, but there are switchers that they want to be activated straight away, so i think it would be good to store that.

src/main/java/com/hedvig/contract/shared/commands/ReviveContract.kt
palmenhq: I don't really understand what you mean here 😅 the definition of a revival is really just adding an agreement after a terminated agreement, so the difference is more symbolic

src/main/java/com/hedvig/contract/valueObjects/CoInsured.kt
palmenhq: I think this is good enough for now, but lets change it as we see the requirements for this. As the member id is optional we decided to allow null here, in order to let the co-insured be "anonymous"

src/main/java/com/hedvig/contract/valueObjects/InsuranceAddOn.kt
palmenhq: Why?

src/main/java/com/hedvig/contract/shared/events/PendingAgreementActivated.kt
```diff
@@ -0,0 +1,11 @@
+package com.hedvig.contract.shared.events
+
+import java.time.LocalDate
+import java.util.UUID
+
+data class PendingAgreementActivated(
```
palmenhq: why is this particular one a data class?

meletios: No idea 😆 we can remove it.

meletios:
Will create another one

PR #345, opened 2019-10-16
Feature/contract agreements db model

meletios:


src/main/java/com/hedvig/contract/apartments/query/dao/ApartmentAgreement.kt
```diff
@@ -0,0 +1,72 @@
...
+  var squareMeters: Long = 0,
+  var removedAt: Instant? = null
+) {
+  fun terminate(terminationDate: LocalDate) {
+    isTerminated = true
+    end = terminationDate
+  }
+
+  fun terminateAfterExpiry(terminationDate: LocalDate) {
+    extendedEnd = terminationDate
```
meletios: Shall we set ' isTerminated = true' or the 'end'

src/main/java/com/hedvig/contract/apartments/query/ApartmentEventListener.kt
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  @EventHandler
+  fun on(
+    e: AgreementRemoved,
+    @Timestamp timestamp: Instant
+  ) {
+    val contract = findContractOrThrow(e.contractId)
+
+    if (contract.pendingAgreement?.id != e.agreementId) {
```
meletios: Oopsie 😆 

```suggestion
    if (contract.pendingAgreement?.id == e.agreementId) {
```

src/main/java/com/hedvig/contract/apartments/query/ApartmentEventListener.kt
```diff
@@ -0,0 +1,206 @@
...
+      contract.pendingAgreement = null
+    }
+
+    findAgreementOrThrow(e.agreementId, contract).removedAt = timestamp
```
meletios: i guess here we throw if the agreement is a pending one? or?

src/main/java/com/hedvig/contract/apartments/query/dao/ApartmentAgreement.kt
meletios: if not? shall this be a one-liner wherever we use it?

palmenhq: I think no because the agreement that carries the extended termination date is technically not terminated right? Its a successor of it that will be terminated.

src/main/java/com/hedvig/contract/apartments/query/ApartmentEventListener.kt
palmenhq: hehe woops

src/main/java/com/hedvig/contract/apartments/query/ApartmentEventListener.kt
palmenhq: ah yes true, the if above should probably not be fall-through

PR #352, opened 2019-10-21
Feature/translate product events to contract

meletios:


src/test/java/com/hedvig/contract/apartments/translators/AgreementSchedulerTest.kt
```diff
@@ -0,0 +1,532 @@
...
+import java.math.BigDecimal
+import java.time.Instant
+import java.time.LocalDateTime
+import java.time.temporal.ChronoUnit
+import java.util.*
+
+class AgreementSchedulerTest {
+
+  @Before
+  fun setUp() {
```
palmenhq: we can kill this now no?

src/test/java/com/hedvig/contract/apartments/translators/AgreementSchedulerTest.kt
```diff
@@ -0,0 +1,532 @@
...
+  }
+
+  @Test
+  fun createAnAgreementFromContractSigned() {
+    val productId1 = PRODUCT_ID_ONE
+    val events = createEvents(
+      MemberCreatedEvent(
+        MEMBER_ID,
+        "191212121212",
+        "Tovlan",
```
palmenhq: `Tolvan`* :)

src/main/java/com/hedvig/contract/apartments/translators/AgreementScheduler.kt
```diff
@@ -0,0 +1,400 @@
...
+  private fun plan(
+    state: MutableList<CompatAgreement>,
+    event: MemberTerminatedEvent,
+    timestamp: Instant
+  ): MutableList<CompatAgreement> {
+    val lastDraftAgreement = state.last()
+
+    val terminateDrafrAgreement = lastDraftAgreement.copy(
+      eventName = event.javaClass.name,
+      createdAt = timestamp,
```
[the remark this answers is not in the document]
meletios: Yes, we create a new row for every event.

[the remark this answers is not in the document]
palmenhq: Ahaa, maybe we could name it something like `timestamp` or `modifiedAt` instead? i just assumed this was the creation date of the agreement

src/main/java/com/hedvig/contract/apartments/translators/AgreementScheduler.kt
```diff
@@ -0,0 +1,396 @@
...
+          }
+          else -> state
+        }
+      }
+  }
+
+  private fun plan(
+    state: MutableList<CompatAgreement>,
+    event: ContractSignedEvent,
+    timeStamp: Instant
```
palmenhq: we're switching a lot between `timeStamp` and `timestamp` and i think we should try to be consistent. Also considering axon framework uses `timestamp` that's what i'd suggest we go with

src/test/java/com/hedvig/contract/apartments/translators/AgreementSchedulerTest.kt
```diff
@@ -0,0 +1,524 @@
...
+
+  @Test
+  fun createAnAgreementFromContractSigned() {
+    val productId1 = PRODUCT_ID_ONE
+    val events = createEvents(
+      MemberCreatedEvent(
+        MEMBER_ID,
+        "191212121212",
+        "Silvio",
+        "Berlusconi",
```
palmenhq: 🙀

src/test/java/com/hedvig/contract/apartments/translators/AgreementSchedulerTest.kt
```diff
@@ -0,0 +1,524 @@
...
+        NOW
+      )
+    )
+
+    val agreementScheduler = AgreementScheduler(events, Blacklist())
+
+    val state = agreementScheduler.generateAgreement(MEMBER_ID)
+
+
+    assertThat(state).hasSize(1)
```
palmenhq: maybe we could do some more granular assertions? To know we actually succeeded

src/test/java/com/hedvig/contract/apartments/translators/AgreementSchedulerTest.kt
```diff
@@ -0,0 +1,524 @@
...
+        25f,
+        Product.ProductTypes.BRF,
+        2,
+        null,
+        null
+      ),
+      ActivationDateUpdatedEvent(
+        MEMBER_ID,
+        productId1,
+        Instant.now()
```
palmenhq: Considering `TODAYS_DATE` is like in the 80's, maybe we should use a timestamp closer to then? Obviously this works but it feels a bit shaky to me. Especially considering we have `NOW` also

src/main/java/com/hedvig/contract/apartments/translators/CompatAgreement.kt
```diff
@@ -0,0 +1,88 @@
...
+  val addressCountry: String = SWEDEN,
+
+  val numberCoInsured: Int? = null,
+  val squareMeters: Long? = null,
+  val currentInsurer: String? = null
+) {
+
+  companion object {
+
+    fun toApartmentAgreement(c: CompatAgreement): ApartmentAgreement {
```
palmenhq: imo this would make more sense to put on the instance, or to put as a `from` on the `ApartmentAgreement`

meletios: Yeap, good point. Will fix

src/main/java/com/hedvig/contract/apartments/translators/AgreementScheduler.kt
```diff
@@ -230,11 +242,15 @@ class AgreementScheduler(
     event: ProductRenewedWithoutReplacementEvent,
     eventTimestamp: Instant
   ): MutableList<CompatAgreement> {
+    if (state.isEmpty()) {
+      throw RuntimeException("Compact agreement state is empty while was trying to plan a ${ProductRenewedWithoutReplacementEvent::class.java.simpleName} event")
```
palmenhq: `Compat`* :)

src/main/java/com/hedvig/contract/apartments/translators/MemberIdContractMap.kt
```diff
@@ -0,0 +1,12 @@
+package com.hedvig.contract.apartments.translators
+
+import java.util.UUID
+import javax.persistence.Entity
+import javax.persistence.Id
+
+@Entity
+data class MemberIdContractMap(
```
palmenhq: i believe `@Entity`s shouldn't be `data` classes :) also isn't this a map entry rather than a map? :) or maybe a mapping?

src/main/java/com/hedvig/contract/apartments/translators/ProductToContractEventListener.kt
```diff
@@ -49,53 +51,19 @@ class ProductToContractEventListener(
       return
     }
 
-    val memberDraftAgreements = handleEvent(e.memberId, timeOfKnowledge)
+    val contractId = (memberIdContractMapRepository.findById(e.memberId).orElseGet {
```
palmenhq: clean!

src/main/java/com/hedvig/contract/apartments/translators/ProductToContractEventListener.kt
```diff
@@ -47,37 +49,145 @@ class ProductToContractEventListener(
     replayStatus: ReplayStatus
   ) {
     if (isBlacklisted(e.memberId, timeOfKnowledge)) {
-      logger.info("Ignoring contract signed event for blacklisted member [memberId={}]", e.memberId)
+      logger.info("Ignoring ${e.javaClass.simpleName} event for blacklisted member [memberId={}]", e.memberId)
```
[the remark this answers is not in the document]
meletios: I think the difference between `e.javaClass.simpleName` and `e::class.java` is that the simplename will return only the name of the class, and the class.java the fullpath + name of the class. Eller? So in order to not have a long log message, dont you think that its better to use the simplename?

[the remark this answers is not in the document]
palmenhq: ah yeah, i see your point. I prefer verbosity over simplicity in most cases, but i guess that's a matter of preference haha :)

src/main/java/com/hedvig/contract/apartments/translators/ProductToContractEventListener.kt
```diff
@@ -47,37 +49,145 @@ class ProductToContractEventListener(
...
+      try {
+        commandBus.sendAndWait<ChangeFromDateOnAgreement>(
+          ChangeFromDateOnAgreement(
+            contractId = contractId,
+            agreementId = draftAgreement.id,
+            newFrom = draftAgreement.from!!
+          )
+        )
+      } catch (e: IllegalArgumentException) {
+        // TODO: Change this, Create its own exception Or check memberDraftAgreements
```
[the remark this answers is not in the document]
meletios: done :D

src/main/java/com/hedvig/contract/apartments/translators/ProductToContractEventListener.kt
meletios: is it clean? or do i need to clean it? 😆

src/test/java/com/hedvig/productPricing/service/web/InternalControllerTest.java
```diff
@@ -144,8 +144,8 @@ public void test2() throws Exception {
         mockMvc.perform(
             get("/_/insurance/searchByActivationDate").param("activationDate", "2018-09-01")
         ).andExpect(status().is2xxSuccessful());
-        assertThat(arg1.getValue().atZone(ZoneId.of("Europe/Stockholm"))).isEqualTo("2018-08-31T22:00:00+02:00");
-        assertThat(arg2.getValue().atZone(ZoneId.of("Europe/Stockholm"))).isEqualTo("2018-09-01T22:00:00+02:00");
+        assertThat(arg1.getValue().atZone(ZoneId.of("Europe/Stockholm"))).isEqualTo("2018-08-31T23:00:00+02:00");
```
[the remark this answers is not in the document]
meletios: I am just a developer 😇 Dont even know why DST or timezones in general exist.

PR #353, opened 2019-10-21
Integrate underwriter with /insurance endpoint

AlexSJ1990:


vonElfvin:
I'm not up to speed with what this is, but shouldn't underwriter have its own endpoints?

src/main/java/com/hedvig/productPricing/service/web/ProductPricingController.java
```diff
@@ -439,8 +444,32 @@ private boolean validateProductTypeUnderwritingLimits(WebOnboardingQuoteRequest
...
+      CompleteQuoteDto quote = this.fakeUnderwriterService.getQuote();
+
       InsuranceDTO insurance = new InsuranceDTO();
       insurance.status = ProductStatus.PENDING;
+
+      if(quote == null) {
+        return ResponseEntity.ok(insurance);
+      }
+
+      if(quote.getCompleteQuoteData() instanceof CompleteQuoteData.CompleteHomeData) {
```
palmenhq: Could we maybe assign `quote.getCompleteQuoteData()` to a variable to not get so many type casts? Gets quite hard to read

src/main/java/com/hedvig/productPricing/campaign/integration/underwriter/dto/CompleteQuoteDto.kt
```diff
@@ -0,0 +1,83 @@
...
+    OBJECT,
+    UNKNOWN
+  }
+
+  enum class LineOfBusiness {
+    BRF,
+    RENT,
+    RENT_BRF,
+    SUBLET_RENTAL,
+    SUBLET_BRF,
```
palmenhq: Are these really needed?

src/main/java/com/hedvig/productPricing/campaign/integration/underwriter/UnderwriterImpl.kt
```diff
@@ -0,0 +1,14 @@
+package com.hedvig.productPricing.campaign.integration.underwriter
+
+import com.hedvig.productPricing.campaign.integration.underwriter.dto.CompleteQuoteDto
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.stereotype.Service
+
+@Service
+class FakeUnderwriterImpl @Autowired constructor(val client: FakeUnderwriterClient): FakeUnderwriter {
```
palmenhq: Actually i think we could let the primary constructor be here without `@Autowire`, spring will detect that its the only constructor :)

src/main/java/com/hedvig/productPricing/campaign/integration/underwriter/Underwriter.kt
```diff
@@ -0,0 +1,8 @@
+package com.hedvig.productPricing.campaign.integration.underwriter
+
+import com.hedvig.productPricing.campaign.integration.underwriter.dto.CompleteQuoteDto
+
+interface FakeUnderwriter {
```
palmenhq: Maybe this should just be `UnderwriterService` or so? So we can choose to either fake the underwriter or actually implement it

src/main/java/com/hedvig/productPricing/campaign/integration/underwriter/UnderwriterImpl.kt
palmenhq: Also is it really a fake underwriter, considering it actually takes a client and calls it? :)

src/main/java/com/hedvig/productPricing/campaign/integration/underwriter/Underwriter.kt
johantjellden: Id go for just calling this interface just `Underwriter` to separate it from "service like classes".

src/main/java/com/hedvig/productPricing/campaign/integration/underwriter/UnderwriterClient.kt
```diff
@@ -0,0 +1,46 @@
...
+        "if",
+        50,
+        2,
+        false,
+        "191212121212",
+        LocalDateTime.now(),
+        listOf()
+      ))
+    }
+}
```
johantjellden: I guess that this one should implement `Underwriter` interface

src/test/java/com/hedvig/productPricing/service/web/ProductPricingControllerTest.kt
```diff
@@ -0,0 +1,128 @@
...
+    Mockito.`when`(underwriter.getQuote(response.memberId!!))
+      .thenReturn(Right(response))
+
+    val request = MockMvcRequestBuilders.get("/insurance").header("hedvig.token", "${response.memberId}")
+
+    val result = mockMvc.perform(request)
+
+    result
+      .andExpect(MockMvcResultMatchers.status().is2xxSuccessful)
+//      .andExpect(MockMvcResultMatchers.jsonPath("$.zipCode").value("11216"))
```
AlexSJ1990: any thoughts on how to fix this? I want to check that the zipCode in the json result response is as expected?

meletios:
@AlexSJ1990 can you check the conflicitng files

src/main/java/com/hedvig/productPricing/campaign/integration/underwriter/dto/QuoteDto.kt
```diff
@@ -0,0 +1,100 @@
...
+
+    val street: String?,
+    val zipCode: String?,
+    val city: String?,
+    var livingSpace: Int?,
+    var householdSize: Int?
+  ): QuoteData()
+}
+
+  enum class QuoteState {
```
palmenhq: this formatting looks a bit off but its a detail :)

src/main/java/com/hedvig/productPricing/service/web/ProductPricingController.java
```diff
@@ -75,7 +77,9 @@ public ProductPricingController(
     RenewalQueueFacade renewalQueueFacade,
     CampaignService campaignService,
     TextKeysLocaleResolver textKeysLocaleResolver,
-    LocalizationService localizationService
+    LocalizationService localizationService,
+    Underwriter underwriterService,
+    com.hedvig.productPricing.service.web.InsuranceEndpointKotlinHelper insuranceEndpointKotlinHelper
```
palmenhq: why do we need the full package name here?

src/main/java/com/hedvig/productPricing/service/web/InsuranceEndpointKotlinHelper.kt
```diff
@@ -0,0 +1,45 @@
+package com.hedvig.productPricing.service.web
```
palmenhq: could we move this to a utils package or so? Considering its not really a web-specific thing

src/main/java/com/hedvig/productPricing/campaign/util/InsuranceEndpointKotlinHelper.kt
```diff
@@ -0,0 +1,45 @@
...
+      } else if (quote.data is QuoteData.ApartmentData) {
+        insurance.addressStreet = quote.data.street
+        insurance.livingSpace = quote.data.livingSpace!!.toFloat()
+        insurance.zipCode = quote.data.zipCode
+        insurance.personsInHousehold = quote.data.householdSize
+      }
+
+      insurance.categories = ArrayList()
+      insurance.newTotalPrice = quote.price!!.toDouble()
+      insurance.currentInsurerName = if (quote.currentInsurer != null) InsuranceCompaniesSE.valueOf(quote.currentInsurer.toUpperCase()) else null
```
palmenhq: is this guy always true? :/ i think we've been here before. What happens if `currentInsurer` is `Länsförsäkringar`? Or is this the same as the enum? Why do we need to uppercase it if so?

PR #354, opened 2019-10-23
No discount discount

FredrikAreschoug:


src/main/java/com/hedvig/productPricing/campaign/web/PartnerCampaignController.kt
```diff
@@ -43,4 +44,16 @@ class PartnerCampaignController(
     return ResponseEntity.accepted().build()
   }
 
+  @PostMapping(value = ["/no-discount/assign"])
```
johantjellden: Should we not put this method on the campaign controller? Since it is quite related to the other method where we assign an campaign with an monthly incentive?

johantjellden: Ah, never mind. Expand the whole files before leaving comments duh.

PR #359, opened 2019-11-05
WIP Add endpoint to replace current free month campaign

vonElfvin:
* Endpoint to replace current active free month campaign with a new one with another code
* Does not support anything else than free months (as of now), thoughts @johantjellden ?


src/main/java/com/hedvig/productPricing/campaign/domain/model/Member.kt
```diff
@@ -134,6 +133,34 @@ open class Member(
     return redeemedCampaign
   }
 
+  fun replaceCurrentRedeemedFreeMonthCampaign(campaign: Campaign) {
+    val currentlyRedeemedCampaign = this.getCurrentRedeemedCampaign()
+        ?: throw RedeemDeclinedException("No currently redeemed campaign to replace", RedeemDeclinedReason.NO_CAMPAIGN_TO_REPLACE)
+
+    if (currentlyRedeemedCampaign.getStatus(YearMonth.now()) != RedeemedCampaignStatus.ACTIVE) {
```
johantjellden: What was the reasoning for this check? I think we talked about it but I don't remeber.

src/main/java/com/hedvig/productPricing/campaign/domain/model/Member.kt
```diff
@@ -134,6 +133,34 @@ open class Member(
...
+
+    val currentActivatedAt = currentlyRedeemedCampaign.activatedAt!!
+    val replacementCampaign = RedeemedCampaign(
+      hibernateId = null,
+      redeemedBy = this,
+      campaign = campaign,
+      redeemedAt = Instant.now(),
+      incentive = campaign.incentive
+    )
+    replacementCampaign.activate(currentActivatedAt)
```
johantjellden: Should we care about dates here? Can we replace a campaign that we have "used" when charging?

src/main/java/com/hedvig/productPricing/campaign/domain/model/Member.kt
vonElfvin: My thought was that we only want to replace campaigns that are active. Not ones in offer for example

src/main/java/com/hedvig/productPricing/campaign/domain/model/Member.kt
vonElfvin: I'm unsure to be honest. I personally feel like it should not matter if it has been used once or not. The use case for this is most commonly if someone has used a code and wants to switch to a better one. Even if the first "weak" code has been used by chance I think they should be allowed to replace it (at least from a technical standpoint).

src/main/java/com/hedvig/productPricing/campaign/domain/model/RedeemedCampaign.kt
```diff
@@ -169,16 +169,16 @@ open class RedeemedCampaign(
     }
   }
 
-  fun setMemberTerminationDate(terminationDate: Instant) {
+  fun terminate(terminateAt: Instant) {
     if (this.unRedeemedAt == null) {
       this.terminatedAt = Instant.now()
-      this.activeTo = terminationDate
+      this.activeTo = terminateAt
```
palmenhq: wat

vonElfvin: What is wat?

src/main/java/com/hedvig/productPricing/campaign/domain/model/Member.kt
johantjellden: Yes, i think it will get quite complicated once we start doing more fancy things. Let's go with this.

src/main/java/com/hedvig/productPricing/campaign/domain/model/RedeemedCampaign.kt
palmenhq: sorry haha 😬

```java
this.terminatedAt = now
this.activeTo = terminateAt
```

seems a bit weird

johantjellden:
I think we should hold of on this, lets go through the usecase again
and see how we should extend campaign-service.

johantjellden:
Lets redo this at a later time.

PR #360, opened 2019-11-08
change the counter to number of insured people with signed insurances

meletios:


src/main/java/com/hedvig/productPricing/service/query/ProductRepository.java
```diff
@@ -66,7 +66,7 @@
     @Param("firstDayOfNextMonth") LocalDateTime firstDayOfNextMonth, @Param("memberId") String memberId
   );
 
-  @Query("select count(*) from ProductEntity p where p.state = 'SIGNED' and (p.activeTo is null or p.activeTo > current_date)")
+  @Query("select count(p.personsInHouseHold) from ProductEntity p where p.state = 'SIGNED' and (p.activeTo is null or p.activeTo > current_date)")
```
vonElfvin: About this query, isn't it missing "where activeFrom IS NOT NULL"

vonElfvin: Also 
```suggestion
  @Query("select sum(p.personsInHouseHold) from ProductEntity p where p.state = 'SIGNED' and (p.activeTo is null or p.activeTo > current_date)")
```

vonElfvin: I.e.

```suggestion
  @Query("select sum(p.personsInHouseHold) from ProductEntity p where p.state = 'SIGNED' and (p.activeTo is null or p.activeTo > current_date) and p.activeFrom IS NOT NULL")
```

akashpatl: nice catch @vonElfvin

meletios:
I am closing this and deleting the branch until we clear the requirements for the counter. 👋

PR #363, opened 2019-11-19
WIP Feature/redeem receipt

johantjellden:
First draft of a flow where we check that a code can be redeem and later add it as a redeemed code to a member.


src/main/java/com/hedvig/productPricing/campaign/domain/model/Incentive.kt
```diff
@@ -49,6 +50,10 @@ data class Incentive private constructor(
       return Incentive(null, amount, currency, IncentiveType.COST_DEDUCTION)
     }
 
+    fun createCostDeduction(ma:MonetaryAmount) : Incentive {
```
meletios: Shall we name it differently? like monetaryAmount?

src/main/java/com/hedvig/productPricing/campaign/service/DomainService.kt
```diff
@@ -138,9 +138,25 @@ class DomainService(
...
 
-    redeemingCampaignOwner.redeemCampaign(campaign, activationDate= activationDate)
+    redeemingMember.redeemCampaign(campaign, activationDate= activationDate)
 
-    campaignOwnerRepository.save(redeemingCampaignOwner)
+    campaignOwnerRepository.save(redeemingMember)
+  }
+
+  fun checkoutRedeemCampaign(campaignCode: String, quoteId: String) : Option<Campaign> {
+    val campaignCodeTrimmedAndLoweCase = CodeUtil.cleanAndLowerCampaignCode(campaignCode, "\$\$underwriter")
```
meletios: Shall we create a constant for `underwriter` and pass the quote id here also?

src/main/java/com/hedvig/productPricing/campaign/service/DomainService.kt
```diff
@@ -138,9 +138,25 @@ class DomainService(
...
+
+    val campaignCodeEntity = campaignCodeRepository.findByIdOrNull(campaignCodeTrimmedAndLoweCase)
+      ?: throw RedeemDeclinedException(
+        "Campaign code $campaignCodeTrimmedAndLoweCase not found",
```
meletios: log the quoteId also here?

src/main/java/com/hedvig/productPricing/campaign/service/DomainService.kt
```diff
@@ -138,9 +138,25 @@ class DomainService(
...
+        RedeemDeclinedReason.CAMPAIGN_CODE_DOES_NOT_EXIST
+      )
+
+    val campaign = campaignRepository.findByIdOrNull(campaignCodeEntity.campaignId)!!
+
+    campaign.validateRedemption(UnsignedMember, Instant.now())
```
meletios: UnsignedMember + quoteId?

src/main/java/com/hedvig/productPricing/campaign/service/DomainService.kt
```diff
@@ -238,11 +253,22 @@ class DomainService(
...
 
-    campaignOwner.setInsuranceActivationDate(insuranceActivatedAt);
+    campaignOwner.setInsuranceActivationDate(insuranceActivatedAt)
 
     campaignOwnerRepository.save(campaignOwner)
   }
 
+  fun assignRedeemedCampaign(memberId: String, campaignId: UUID, incentive: Incentive) {
+    val member = campaignOwnerRepository.findByMemberId(memberId)
+    require(member != null)
```
meletios: Shall we throw a custom exception here?

src/main/java/com/hedvig/productPricing/campaign/service/DomainService.kt
```diff
@@ -238,11 +253,22 @@ class DomainService(
...
+
+    val campaign = campaignRepository.findByIdOrNull(campaignId)
+    require(campaign != null)
```
meletios: Shall we throw a custom exception here?

src/main/java/com/hedvig/productPricing/campaign/web/MemberCampaignController.kt
```diff
@@ -57,4 +59,34 @@ class MemberCampaignController(
...
 }
+
+data class AssignRedeemedCodeRequest(
+  val memberId: String,
+  val campaignId: UUID,
+  val incentive: Incentive
+  )
+
+data class CheckoutRedeemResponse(
+  //TODO since Incentive belongs to graphql namespace we need to write tests in controller that verifies that we
```
meletios: Shall we use different classes in order to not break something accidentally?

src/main/java/com/hedvig/productPricing/campaign/service/DomainService.kt
meletios: This will fail if the campaigns are referrals. So i think, we should limit to vouchers (not sure what is the business case but i suppose is only for vouchers)

src/main/java/com/hedvig/productPricing/campaign/service/DomainService.kt
```diff
@@ -138,9 +138,25 @@ class DomainService(
...
     val campaign = campaignRepository.findByIdOrNull(campaignCodeEntity.campaignId)!!
 
-    redeemingCampaignOwner.redeemCampaign(campaign, activationDate= activationDate)
+    redeemingMember.redeemCampaign(campaign, activationDate= activationDate)
 
-    campaignOwnerRepository.save(redeemingCampaignOwner)
+    campaignOwnerRepository.save(redeemingMember)
+  }
+
+  fun checkoutRedeemCampaign(campaignCode: String, quoteId: String) : Option<Campaign> {
```
meletios: Shall we rename this to validatePossibleRedemption or something else? 😆

src/main/java/com/hedvig/productPricing/campaign/web/MemberCampaignController.kt
```diff
@@ -57,4 +59,34 @@ class MemberCampaignController(
       MDC.remove("memberId")
     }
   }
+
+  @PostMapping("/checkoutRedeem")
```
meletios: Shall we rename this to validatePossibleRedemption or something else? 😆

src/main/java/com/hedvig/productPricing/campaign/domain/model/Incentive.kt
palmenhq: what about `amountToDeduct`? or something descriptive. I can see that its a monetary amount already

src/main/java/com/hedvig/productPricing/campaign/service/DomainService.kt
palmenhq: I'm not really a fan of this hack, maybe we could overload the `validateRedemption` for non-target-member campaigns

src/main/java/com/hedvig/productPricing/campaign/domain/model/Member.kt
```diff
@@ -281,5 +285,13 @@ open class Member(
       return null
     }
   }
+
+  fun forceRedeemCampaign(campaign: Campaign, incentive: Incentive) {
+    if(this.getCurrentRedeemedCampaign() != null){
+      throw RuntimeException("Cannot force redeem a campaign when there already exists an existing redeemed campaign")
```
palmenhq: is this always true?

src/main/java/com/hedvig/productPricing/campaign/service/DomainService.kt
johantjellden: Yes maybe but I dont know if it is really hacky, The problem is that further down the line we will do some validation on the member tries to redeem a code. We could refactor all that logic checks to see for instance if we pass in a null instead of a member. But that is even more hacky.

The `UnsignedMember`is its own type and not just a constant. I can be extended to other special cases in the future. 

Our usecase with an signed and existing member that tries to redeem a campaign is still valid and will be even when we only use underwriter for all quotes.

src/main/java/com/hedvig/productPricing/campaign/domain/model/Member.kt
johantjellden: Yes for this usecase, maybe we need to change the name of the method. But if underwriter tries to add two redeemed campaigns to a member I think it is best for us to deny that.

src/main/java/com/hedvig/productPricing/campaign/service/DomainService.kt
palmenhq: Imo `null` would be a better option? It's literally the representation of that there's nothing there - and not having a member would be the perfect match for that case. Especially considering this is Kotlin and it has null safety checks. When looking up the data i would rather search for `null` than a "magic" member id.

johantjellden:
I think its time to close this.

PR #364, opened 2019-11-20
use graphql commons

FredrikAreschoug:
Did this because I needed to test it in another service than `Underwriter`


FredrikAreschoug:
the package is not public for some reason looking into it

FredrikAreschoug:
Unfortunately, we need to add an authentication token to install public github package and I felt that that would be to much hassle to do for everyone so changed to bintray.

src/main/java/com/hedvig/productPricing/campaign/graphql/types/Incentive.kt
```diff
@@ -1,10 +1,11 @@
 package com.hedvig.productPricing.campaign.graphql.types
 
 import com.hedvig.productPricing.campaign.domain.model.IncentiveType
+import com.hedvig.type.MonetaryAmountV2
```
palmenhq: i haven't really been part of making this package but this feels like it should be part of a more granular subpackage, something like `com.hedvig.graphl.types.MonetaryAmountV2` or something?

PR #362, opened 2019-11-18
Contract/Agreements

meletios:

PR #365, opened 2019-11-20
Feature/contract agreements validator

meletios:


src/main/java/com/hedvig/contract/apartments/query/ApartmentEventListener.kt
```diff
@@ -0,0 +1,268 @@
...
+  @EventHandler
+  fun on(e: ContractWithActiveAgreementCreated<ApartmentAgreement>, @Timestamp timestamp: Instant) {
+    logger.info("[${e.javaClass.simpleName}] Handling event for [memberId={${e.holderMemberId}]")
+    val newContract = ApartmentContractEntity
+      .fromApartmentContractWithActiveAgreement(
+        contractId = e.contractId,
+        holderMemberId = e.holderMemberId,
+        apartmentAgreement = e.agreement,
+        switchedFrom = e.switchedFrom,
+        createdAt = timestamp
```
palmenhq: what about the original timestamp, don't we need to use it?

src/main/java/com/hedvig/contract/apartments/translators/AgreementScheduler.kt
```diff
@@ -517,4 +521,13 @@ class AgreementScheduler(
         !(compatAgreement.payload as ProductModifiedEvent).terminationDate.isBefore(terminationDate) &&
         !compatAgreement.timestamp.isBefore(cancellationEventTimestamp)
     }
+
+  private fun isTerminatePreviously(
```
palmenhq: isTerminate**d**Previously?

src/main/java/com/hedvig/contract/apartments/translators/query/ErrorProduct.kt
```diff
@@ -0,0 +1,20 @@
+package com.hedvig.contract.apartments.translators.query
+
+import java.time.Instant
+import javax.persistence.Entity
+import javax.persistence.Id
+import org.hibernate.annotations.CreationTimestamp
+import org.hibernate.annotations.UpdateTimestamp
+
+@Entity
+data class ErrorProduct(
```
palmenhq: `ErrorProduct`? Technically this isn't really a product but rather an `ErrorEvent` or member or something no?

src/main/java/com/hedvig/contract/apartments/translators/query/ErrorProduct.kt
```diff
@@ -0,0 +1,20 @@
...
+  @Id
+  val memberId: String,
+  val eventName: String?,
+  val exceptionMessage: String?
+) {
+  @CreationTimestamp
+  lateinit var createdAt: Instant
+  @UpdateTimestamp
```
palmenhq: imo we shouldn't update but rather make a new post and find the maximum, but i guess it's _fine_ in this particular case

src/main/java/com/hedvig/contract/apartments/validator/ContractProductValidator.kt
```diff
@@ -0,0 +1,89 @@
...
+    val now = LocalDate.now()
+    val latestAgreement = if (contract.agreements.isNotEmpty()) contract.agreements
+      .filter { apartmentAgreement -> !apartmentAgreement.from!!.isAfter(now) }
+      .maxBy { apartmentAgreement -> apartmentAgreement.from!! }!! else contract.pendingAgreement!!
+
+    val productAddress = if (product.address != null) product.address else product.member.address
+
+    require(product.id == latestAgreement.id) { "[id] [Product] ${product.id} [Agreement] ${latestAgreement.id}" }
+
+    require(product.activeFrom.toLocalDate() == latestAgreement.from) { "[From] [Product] ${product.activeFrom} [Agreement] ${latestAgreement.from}" }
```
palmenhq: just a thought shall we maybe store all the reasons instead of one at a time?

src/main/java/com/hedvig/contract/apartments/translators/query/ErrorProduct.kt
meletios: This is fixed in another PR https://github.com/HedvigInsurance/product-pricing/pull/362

src/main/java/com/hedvig/contract/apartments/validator/ContractProductValidator.kt
meletios: Good point! Will fix it

src/main/java/com/hedvig/contract/apartments/query/ApartmentEventListener.kt
meletios: Oh yeah, thats correct! Will fix dat

src/main/java/com/hedvig/contract/apartments/validator/ContractProductValidator.kt
```diff
@@ -0,0 +1,103 @@
...
+        contractProductValidationOutputRepository.save(contractProductValidationOutput)
+      }
+    }
+  }
+
+  fun getValidationResult(): MutableList<ContractProductValidationOutput> {
+    return contractProductValidationOutputRepository.findAll()
+  }
+
+  private fun areSame(
```
palmenhq: i think this should rather be a list of differences, and we should rename this method to something more matching what is does, like `findDifferences(...): List<String>` or so. Then we could always check if the list is empty

meletios: Nice idea! Just fixed it

src/main/java/com/hedvig/contract/apartments/query/ApartmentEventListener.kt
```diff
@@ -0,0 +1,289 @@
...
+    val contract = findContractOrThrow(e.contractId)
+
+    logger.info("[${e.javaClass.simpleName}] Handling event for [memberId={${contract.holderMemberId}]")
+
+    findAgreementOrThrow(e.agreementId, contract).terminateAfterExpiry(e.terminationDate)
+    apartmentContractRepository.save(contract)
+  }
+
+  @EventHandler
+  fun on(e: ContractRevived<ApartmentAgreement>, @Timestamp timestamp: Instant) {
```
palmenhq: we don't have an original timestamp here or?

src/main/java/com/hedvig/contract/apartments/query/ApartmentEventListener.kt
```diff
@@ -0,0 +1,289 @@
...
+      return
+    }
+
+    contract.agreements.remove(findAgreementOrThrow(e.agreementId, contract))
+
+    apartmentContractRepository.save(contract)
+  }
+
+  @EventHandler
+  fun on(e: AgreementRenewed<ApartmentAgreement>, @Timestamp timestamp: Instant) {
```
palmenhq: Same, no original timestamp?

PR #369, opened 2019-11-29
If no member found when redeeming then create memeber

johantjellden:
----

Currently we fail about every 10th redeem from underwriter due to missing member


src/test/java/com/hedvig/productPricing/campaign/domain/DomainServiceTest.kt
```diff
@@ -133,7 +133,7 @@ class DomainServiceTest {
   }
 
   @Test
-  @Ignore("need to test in DEV that this works, remove test in future")
+  //@Ignore("need to test in DEV that this works, remove test in future")
```
johantjellden:
```suggestion
```

PR #370, opened 2019-11-29
Manual correction endpoint

meletios:
Add manual enpoints to handle products which we cannot translate to Contracts


src/main/java/com/hedvig/contract/apartments/ApartmentAgreement.kt
```diff
@@ -111,5 +112,42 @@ data class ApartmentAgreement(
...
+        from = dto.from,
+        to = dto.to,
+        address = Address(
+          dto.addressStreet,
+          dto.addressCoLine,
+          dto.addressPostalCode,
+          dto.addressCity,
+          dto.addressCountry
+        ),
+        coInsured = if (dto.numberCoInsured - 1 > 0) {
```
vonElfvin: Does the numberCoInsured include the one person insured? Shouldn't it be 0 or null if there are no consinured?

```suggestion
        coInsured = if (dto.numberCoInsured - 1 > 0) {
```

src/main/java/com/hedvig/contract/apartments/ApartmentAgreement.kt
```diff
@@ -111,5 +112,42 @@ data class ApartmentAgreement(
...
+          MutableList(dto.numberCoInsured - 1) { CoInsured() }
+        } else {
+          emptyList()
+        },
+        squareMeters = dto.squareMeters,
+        termsAndConditions = "", // TODO: WHOAT
```
vonElfvin: Indeed

PR #371, opened 2019-12-02
Add datadog agent

johantjellden:


johantjellden:
We need to deploy this, should not break anything I will use **admin merge power**.

PR #372, opened 2019-12-02
Feature/improved search

palmenhq:
extends https://github.com/HedvigInsurance/member-registry-service/pull/341


src/main/java/com/hedvig/productPricing/service/web/InternalController.java
```diff
@@ -335,6 +336,15 @@ public ProductSearchResultDTO searchPaged(ProductSearchRequestDTO searchRequest)
     return res;
   }
 
+  @RequestMapping(value = "/extendMemberSearch", method = RequestMethod.POST)
+  public ResponseEntity<List<MemberSearchResultDTOExtended>> extendMemberSearch(@RequestBody List<String> memberIds) {
+    if (memberIds.size() > 500) {
```
FredrikAreschoug: 500 without paging? :)

palmenhq: Just wanted to make _some_ kind of sanity check if for example the paging would fail

PR #373, opened 2019-12-02
Feature/contract own blacklist

meletios:
Use own blacklist for contract-aggreements


src/main/java/com/hedvig/contract/util/ContractBlacklist.kt
```diff
@@ -0,0 +1,267 @@
...
+    Pair(
+      "297960483",
+      BlacklistItem(
+        BlacklistReason.INVALID_RENEWAL
+      )
+    ),
+    Pair(
+      "159827849",
+      BlacklistItem(
+        // /WTF ask IEX
```
vonElfvin: Lol

PR #376, opened 2019-12-02
Feature/contract agreement power of attorney

meletios:
_First check https://github.com/HedvigInsurance/product-pricing/pull/373 (Contract blacklist)_

Add power of attorney bytearray to contracts


src/main/java/com/hedvig/productPricing/service/events/InsuranceMandateCreatedEvent.java
```diff
@@ -1,14 +1,43 @@
 package com.hedvig.productPricing.service.events;
 
-import lombok.Value;
+public final class InsuranceMandateCreatedEvent {
+  private final String memberId;
+  private final byte[] mandatePdf;
```
vonElfvin: You could just make these public so you don't need all the "delombok" code below

src/main/java/com/hedvig/contract/util/Country.kt
```diff
@@ -0,0 +1,5 @@
+package com.hedvig.contract.util
+
+enum class Country {
+  SWEDEN
+}
```
vonElfvin: I think we should use a package for this so we can seemlessly handle multiple types of country representations (country code etc.)  I did some digging before and this one seems like the best: https://github.com/TakahikoKawasaki/nv-i18n

meletios:
> We may need to discuss this with Fredrik and Nordlin, implementationwise it looks good though

Correct, i will not merge this until i have the discussions with Fredrik and Karl

meletios:
Will be stored in the member-service.

PR #380, opened 2019-12-04
underwriter signed quote endpoint

FredrikAreschoug:


src/main/java/com/hedvig/productPricing/campaign/integration/underwriter/dto/QuoteDto.kt
```diff
@@ -51,6 +55,7 @@ sealed class QuoteData {
     val street: String? = null,
     val city: String? = null,
     val zipCode: String? = null,
+    val floor: Int? = null,
```
vonElfvin: This is never used

src/main/java/com/hedvig/productPricing/service/commands/SingedContractCommand.java
```diff
@@ -21,4 +20,56 @@
     private final LocalDateTime activeFrom;
     private final Boolean isFromPartner;
 
+    public SingedContractCommand(
```
vonElfvin: Singed is a good champion in League of Legends

src/main/java/com/hedvig/productPricing/service/commands/SingedContractCommand.java
```diff
@@ -7,7 +7,6 @@
 import java.time.Instant;
 import java.time.LocalDateTime;
 
-@Value
```
vonElfvin: You can just make the variables public so you don't need to to delombok the whole file

src/main/java/com/hedvig/productPricing/service/serviceIntegration/botService/dto/Address.java
```diff
@@ -9,15 +9,22 @@
...
   private String zipCode = "";
   private Integer floor = 0;
 
+  public Address() {}
+
+  public Address(String street, String city, String zipCode, Integer floor) {
+     this.street = street;
+     this.city = city;
+     this.zipCode = zipCode;
+     this.floor = floor;
```
vonElfvin: floor is never used

src/main/java/com/hedvig/productPricing/service/commands/SingedContractCommand.java
meletios: Really good  tanky player, i would add.

PR #383, opened 2019-12-05
eligibleToSign

FredrikAreschoug:


src/main/java/com/hedvig/productPricing/service/web/InternalController.java
```diff
@@ -787,7 +769,12 @@ public ProductSearchResultDTO searchPaged(ProductSearchRequestDTO searchRequest)
 
     Optional<ProductEntity> product = insuranceService.getCurrentOrLastInsurance(memberId);
     if (!product.isPresent()) {
-      return ResponseEntity.notFound().build();
+      InsuranceDTO insuranceDTO = insuranceEndpointKotlinHelper.getInsurance(memberId);
+      if (insuranceDTO.cost != null) {
```
johantjellden: Why do we filter away insurances with 0 cost?

PR #388, opened 2019-12-09
Feature/percentage discount

akashpatl:
Support Campaign vouchers where we give percentage discount for a fixed period of time


src/main/java/com/hedvig/productPricing/campaign/web/PartnerCampaignController.kt
```diff
@@ -31,9 +30,21 @@ class PartnerCampaignController(
 
     return ResponseEntity.accepted().build()
   }
-
+  //TODO: change mapping name to /voucher-free-moths/ ?
```
FredrikAreschoug: Do it, I'm pretty sure this is only used when we create it with port forwarding into the pod :)

src/main/java/com/hedvig/productPricing/campaign/domain/model/Incentive.kt
```diff
@@ -46,19 +48,46 @@ data class Incentive private constructor(
...
+      return Incentive(
+        numberOfMonths = nrFreeMonths,
+        percentageDiscount = BigDecimal.valueOf(100),
+        amount = null,
+        currency = null,
+        type = IncentiveType.FREE_MONTHS
+      )
     }
 
+    fun createPercentageMonthDiscount(nrMonths: Int, percentageDiscount: BigDecimal?): Incentive {
```
meletios: Should we force percentageDiscount parameter to be non-nullable here?

src/main/java/com/hedvig/productPricing/campaign/domain/model/Incentive.kt
```diff
@@ -12,8 +13,9 @@ import javax.persistence.Enumerated
 
 @Embeddable
 data class Incentive private constructor(
-  val numberOfFreeMonths: Int?,
+  val numberOfMonths: Int?,
   val amount: BigDecimal?,
+  val percentageDiscount: BigDecimal? = BigDecimal.valueOf(0),
```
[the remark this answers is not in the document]
meletios: i noticed that we don't initiate all the other variables. 

Shall we initiate them or shall we not initiate percentageDiscount in order to keep the same "logic" for all the variables?

src/main/java/com/hedvig/productPricing/campaign/domain/model/Member.kt
```diff
@@ -253,33 +255,34 @@ open class Member(
...
-      )
-    }
+      val redeemedCampaign = redeemedCampaigns.firstOrNull {
+        it.campaign is VoucherCampaign && it.incentive.type == incentiveType
+          && it.shouldProvideDiscount(
+          discountChecker = this,
+          monthOfCheck = month,
+          billingDiscount = true
+        )
+      }
```
meletios: White spaceeees 🏳

src/main/java/com/hedvig/productPricing/campaign/service/CampaignBillingServiceImpl.kt
```diff
@@ -88,30 +89,45 @@ class CampaignBillingServiceImpl(
...
+    } else {
+
+      val totalVoucherPercentageDiscount = getTotalVoucherDiscount(member, month, data, referralDiscountAmount, IncentiveType.MONTHLY_PERCENTAGE_DISCOUNT_FIXED_PERIOD)
+
+      if (totalVoucherPercentageDiscount != null) {
+        listOfCampaignDiscount +=
+          CampaignDiscount(
+            memberId,
+            totalVoucherPercentageDiscount.first.with(Monetary.getRounding(RoundingQueryBuilder.of().setScale(0).set(RoundingMode.HALF_EVEN).build())),
+            DiscountType.PERCENTAGE_DISCOUNT,
```
meletios: This one goes to account-service right? 
Dont forget to change account-service 😁

src/main/java/com/hedvig/productPricing/campaign/service/CampaignBillingServiceImpl.kt
```diff
@@ -88,30 +89,45 @@ class CampaignBillingServiceImpl(
...
         CampaignDiscount(
           memberId,
-          totalVoucherDiscount.first.with(Monetary.getRounding(RoundingQueryBuilder.of().setScale(0).set(RoundingMode.HALF_EVEN).build())),
+          totalVoucherFreeMonthDiscount.first.with(Monetary.getRounding(RoundingQueryBuilder.of().setScale(0).set(RoundingMode.HALF_EVEN).build())),
           DiscountType.FREE_MONTH,
-          totalVoucherDiscount.second
+          totalVoucherFreeMonthDiscount.second
         )
-    }
+    } else {
```
meletios: instead of the else statement, is it better if we return the listOfCampaignDiscount ?

src/main/java/com/hedvig/productPricing/campaign/domain/model/Incentive.kt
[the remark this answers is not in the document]
akashpatl: Good point. I don't think we should initiate it there but rather let the creation of the instance set it to null or a value, like all other variables.

src/main/java/com/hedvig/productPricing/campaign/domain/model/Incentive.kt
akashpatl: yes!

src/main/java/com/hedvig/productPricing/campaign/domain/model/Member.kt
akashpatl: 😱

src/main/java/com/hedvig/productPricing/campaign/service/CampaignBillingServiceImpl.kt
akashpatl: Hmm yes, maybe that would be clearer. The you would need a follow up check that sees if any of the campaigns are free months, if so return list else get PercentageDiscount campaigns

src/main/java/com/hedvig/productPricing/campaign/service/CampaignBillingServiceImpl.kt
akashpatl: Did not know that! Thanks!

src/main/java/com/hedvig/productPricing/campaign/config/GraphQLConfiguration.kt
```diff
@@ -29,6 +29,7 @@ class GraphQLConfiguration {
           Referral.AcceptedReferral::class.java,
           Referral.TerminatedReferral::class.java,
           Incentive.FreeMonths::class.java,
+          Incentive.PercentageDiscountMonths::class.java,
```
johantjellden: What  effect does a new incentive have on the app and web-onboarding? @hedvigoscar @sampettersson

hedvigoscar: The app will not do anything when encountering an unknown type of incentive, essentially

hedvigoscar: Also, we should start thinking about the concept of the apps, not a singular app - we have two and may often act in different ways

src/main/java/com/hedvig/productPricing/campaign/domain/model/Member.kt
```diff
@@ -228,16 +228,18 @@ open class Member(
...
+      percentageDiscount: BigDecimal,
       redeemedCampaignActiveAt: LocalDate,
       redeemedCampaignEndDate: LocalDate?): MonetaryAmount? {
+
       return billablePeriods
         .map { billablePeriod ->
           val elapsedDays = ChronoUnit.DAYS.between(billablePeriod.activeFrom, billablePeriod.activeTo) + 1
-          val pricePerDay = billablePeriod.subscription.divide(elapsedDays)
-
+          val percentageDiscountInDecimal =  percentageDiscount.divide(BigDecimal(100)).setScale(4, RoundingMode.HALF_EVEN)
```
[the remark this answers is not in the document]
akashpatl: good point! Shouldn't even need a rounding here. It's done later anyway.

src/main/java/com/hedvig/productPricing/campaign/domain/model/Incentive.kt
```diff
@@ -12,8 +13,9 @@ import javax.persistence.Enumerated
 
 @Embeddable
 data class Incentive private constructor(
-  val numberOfFreeMonths: Int?,
+  val numberOfMonths: Int?,
```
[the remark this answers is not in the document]
akashpatl: is the above annotation enough as a cirumvention of migration @johantjellden?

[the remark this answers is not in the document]
johantjellden: Yes I think so

src/main/java/com/hedvig/productPricing/campaign/config/GraphQLConfiguration.kt
akashpatl: ok! So we'll need to add support in the apps for this discount type take affect. Noted!

src/main/java/com/hedvig/productPricing/campaign/graphql/types/InsuranceCost.kt
```diff
@@ -3,7 +3,7 @@ package com.hedvig.productPricing.campaign.graphql.types
 import com.hedvig.graphql.commons.type.MonetaryAmountV2
 import java.time.LocalDate
 
-data class InsuranceCost(
+data class  InsuranceCost(
```
meletios: White space 😇

hedvigoscar: Friendly recommendation: Add `ktlint` (https://github.com/pinterest/ktlint) to your editors, then add `Save Actions` and have it run code formatting when you save files. This will save you from this category of PR issues forever

johantjellden:
I think we can merge this into master and then make sure not to create any live campaigns before we have updated the clients.

PR #391, opened 2019-12-11
fix status for terminated insurances at the same day

meletios:


akashpatl:
Will this finally make it so that when `active_from == active_to` (but not null) the state is considered Terminated? :)

johantjellden:
What is this pull request for?

PR #397, opened 2019-12-12
Add explicit exceptions to contract agreements

vonElfvin:


src/main/java/com/hedvig/contract/exceptions/ContractAlreadyCreatedException.kt
```diff
@@ -0,0 +1,37 @@
...
+  companion object {
+    fun withPendingAgreement(
+      contract: BaseContract<BaseAgreement>,
+      command: Any,
+      agreementId: UUID
+    ): ContractAlreadyCreatedException {
+      return ContractAlreadyCreatedException(
+        contract = contract,
+        command = command,
+        message = "A pending agreement with id=$agreementId already exists"
```
meletios:
```suggestion
        message = "Contract with a pending agreement with id=$agreementId already exists"
```

PR #398, opened 2019-12-13
calculate incurence cost

FredrikAreschoug:


src/main/java/com/hedvig/productPricing/service/web/UnderwriterController.kt
```diff
@@ -147,6 +155,15 @@ class UnderwriterController(
     )
   }
 
+  @PostMapping("/{memberId}/calculate/insurance/cost")
+  private fun getInsuranceCostOnPrice(
```
johantjellden: This should probably be a public method. https://stackoverflow.com/questions/50563524/requestmapping-works-on-private-methods

PR #399, opened 2019-12-13
Add house to translation+validation, trim zip codes in translation 

vonElfvin:


src/main/java/com/hedvig/contract/translators/ProductToContractEventListener.kt
```diff
@@ -197,10 +199,10 @@ class ProductToContractEventListener(
             )
           )
         }
-        commandBus.sendAndWait<AddAgreement<ApartmentAgreement>>(
+        commandBus.sendAndWait<AddAgreement<BaseAgreement>>(
```
meletios:
```suggestion
        commandBus.sendAndWait<Void>(
```

meletios: is referring to the return type, which is actually void

src/main/java/com/hedvig/productPricing/service/aggregates/MemberAggregate.java
```diff
@@ -1,12 +1,12 @@
 package com.hedvig.productPricing.service.aggregates;
 
+import com.hedvig.contract.shared.models.objects.ExtraBuilding;
```
meletios: Do we use this?

src/main/java/com/hedvig/productPricing/service/aggregates/Product.java
```diff
@@ -1,5 +1,6 @@
 package com.hedvig.productPricing.service.aggregates;
 
+import com.hedvig.contract.shared.models.objects.ExtraBuilding;
```
meletios: Do we use this?

vonElfvin: Yes, since "ExtraBuilding" is identical now from before I just moved it to contracts for the feeling of it being in Contracts from now on

PR #400, opened 2019-12-13
update exception on timeline

meletios:


src/main/java/com/hedvig/contract/shared/timeline/exceptions/TimelineTerminatedException.kt
```diff
@@ -1,10 +1,6 @@
...
-    fun terminated(): TimelineTerminatedException =
-      TimelineTerminatedException(
-        "Timeline terminated when it should not"
-      )
-  }
-}
+import com.hedvig.contract.shared.timeline.TimelinePart
+
+class TimelineTerminatedException(timelinePart: TimelinePart) :
+  RuntimeException("Timeline terminated when it should not [TimelinePartId: ${timelinePart.id}]")
```
vonElfvin: Would be nice to use a TimelineException which always prints the ID of the timeline regardless of message. Where TimelineTerminatedException extends TimelineException which in turn extends RuntimeExceptions. That way we never miss logging the ID of the timeline

PR #414, opened 2020-01-06
Handle renewals and certificate generation completely in contract agreements

vonElfvin:


src/main/java/com/hedvig/productPricing/service/web/InternalRenewalController.kt
```diff
@@ -17,71 +19,26 @@ import org.springframework.web.bind.annotation.RestController
...
-    if (renewalFacade.isProductQueuedForRenewal(productToRenewId)) return
-    if (renewalFacade.hasMultipleNonRenewedPolicies(policyToRenew.memberId)) return
-
-    val renewalDate = policyToRenew.expiryDateInStockholm.plusDays(1)
-    commandGateway.sendAndWait<Void>(InitiateRenewalOfProductCommand(
-      policyToRenew.memberId,
-      productToRenewId,
-      renewalDate
-    ))
+  fun getActiveRenewals(): ResponseEntity<List<RenewalState>> {
```
AlexSJ1990: shouldn't this return ```ResponseEntity<Void>```

PR #416, opened 2020-01-07
fix/percentage discount

FredrikAreschoug:
Fix for percentage discount type. The new code was not back-compatible with people who already have utilized a free month campaign. I've changed the FREE_MONTH IncentiveType to not have a default percentage discount of 100%. Rather this will be accounted for when the actual discount calculation is made instead.


src/main/java/com/hedvig/productPricing/campaign/domain/model/Incentive.kt
```diff
@@ -2,18 +2,22 @@ package com.hedvig.productPricing.campaign.domain.model
 
 import com.hedvig.productPricing.campaign.util.DateUtil
 import com.hedvig.productPricing.util.ProRataCalculator
+import jdk.jfr.Percentage
```
palmenhq: is this intentional?

src/main/java/com/hedvig/productPricing/campaign/domain/model/Member.kt
```diff
@@ -253,33 +256,41 @@ open class Member(
...
-      if (remainingToZero == null || remainingToZero <= Money.of(0, "SEK"))
+    if (redeemedCampaign != null) {
+
+      val percentageDiscount =
+        if (incentiveType.equals(IncentiveType.FREE_MONTHS))
+          BigDecimal.valueOf(100)
+        else if (incentiveType.equals(IncentiveType.MONTHLY_PERCENTAGE_DISCOUNT_FIXED_PERIOD))
+          redeemedCampaign.campaign.incentive.percentageDiscount!!
+        else
+          throw NullPointerException("Wrong incentive type with a null percentage discount")
```
palmenhq: a) i don't think this should be an NPE, a `RuntimeExecption` (or named exception) would be better imo
b) could we add the actual incentive type to the error message? Imagine looking at this in sentry being like "ok, now what?"

src/main/java/com/hedvig/productPricing/campaign/domain/model/Member.kt
```diff
@@ -253,33 +256,41 @@ open class Member(
...
-        month,
-        billablePeriods,
-        DateUtil.toLocalDate(redeemedCampaign.activatedAt!!),
-        redeemedCampaign.endDateOfDiscount()
-      )
-      if (remainingToZero == null || remainingToZero <= Money.of(0, "SEK"))
+    if (redeemedCampaign != null) {
+
+      val percentageDiscount =
+        if (incentiveType.equals(IncentiveType.FREE_MONTHS))
```
palmenhq: kotlin translates `==` to `.equals` (or something similar in bytecode at least) i think, so there's no need to use `.equals` here is there? :)

src/main/java/com/hedvig/productPricing/campaign/domain/model/Member.kt
```diff
@@ -253,33 +256,41 @@ open class Member(
...
+      }
 
-      val remainingToZero = calculateFreeMonthsDiscountByActiveDays(
-        month,
-        billablePeriods,
-        DateUtil.toLocalDate(redeemedCampaign.activatedAt!!),
-        redeemedCampaign.endDateOfDiscount()
-      )
-      if (remainingToZero == null || remainingToZero <= Money.of(0, "SEK"))
+    if (redeemedCampaign != null) {
```
palmenhq: maybe we could negate this if and return early? To not have so many nested ifs

src/main/java/com/hedvig/productPricing/campaign/domain/model/Member.kt
```diff
@@ -253,33 +256,41 @@ open class Member(
...
+          throw NullPointerException("Wrong incentive type with a null percentage discount")
+
+        val remainingToZero = calculatePercentageDiscountByActiveDays(
+          month,
+          billablePeriods,
+          percentageDiscount,
+          DateUtil.toLocalDate(redeemedCampaign.activatedAt!!),
+          redeemedCampaign.endDateOfDiscount()
+        )
+        if (remainingToZero == null || remainingToZero <= Money.of(0, "SEK"))
```
palmenhq: could we add curly brackets here? I think its too easy to make fall-through statements without them

PR #420, opened 2020-01-09
Feature/contract agreements facade safety increasers

vonElfvin:


vonElfvin:
Solved with different approach

PR #421, opened 2020-01-09
Feature/contract agreements facade search

vonElfvin:


vonElfvin:
Solved with different solution

PR #424, opened 2020-01-10
update-manual-contract-endpoints

meletios:
add two new manual endpoints for contacts


src/main/java/com/hedvig/contract/service/ManualCorrectionServiceImpl.kt
```diff
@@ -106,4 +111,35 @@ class ManualCorrectionServiceImpl(
       )
     )
   }
+
+  override fun changeToDate(req: ManualToDateRequest) {
+    commandGateway.sendAndWait<Void>(
+      ChangeToDateOnAgreement(
+        contractId = contractStateRepository.findByHolderMemberId(req.memberId).firstOrNull()!!.id,
```
FredrikAreschoug: why not just `first()` when you are using `!!`?

meletios: good catch! 👍

PR #427, opened 2020-01-10
WIP Contract agreement takeover

vonElfvin:


src/main/java/com/hedvig/productPricing/service/query/ContractEntity.java
```diff
@@ -3,7 +3,7 @@
 import javax.persistence.Entity;
 import javax.persistence.Id;
 
-
+// TODO: Fix for C&A Write: Ensure these are still created in the future
```
meletios: Do we even need this entity? or should we kill with 🔥 ?

PR #434, opened 2020-01-21
wip create contract from underwriter

AlexSJ1990:


vonElfvin:
Similar idea is now done via Norway implementation

PR #435, opened 2020-01-22
Feature/mandate

meletios:
- New aggregate for mandates
- Uses the same contract repository
- Generates the "new" mandates when a quote is signed from underwriter


src/main/java/com/hedvig/productPricing/service/web/dto/InsuranceDTO.java
```diff
@@ -116,6 +117,23 @@ public InsuranceDTO(Clock clock, ProductEntity product, List<PerilEntity> perils
                             : null;
     this.isSubleted = product.isSubleted;
 
+    Integer personInHousehold = null;
+    if (product.personsInHouseHold != null){
+      personInHousehold = product.personsInHouseHold;
+    } else if (ue != null && ue.personsInHouseHold != null) {
```
AlexSJ1990: can we use a better name that ue?

PR #447, opened 2020-01-28
Include generated certificate urls when fetching a member's insurances

vonElfvin:


src/main/java/com/hedvig/productPricing/service/web/InternalController.java
```diff
@@ -309,7 +311,10 @@ public ProductSearchResultDTO searchPaged(ProductSearchRequestDTO searchRequest)
     }
 
     List<InsuranceStatusDTO> insurances =
-      products.stream().map(InsuranceStatusDTO::new).collect(Collectors.toList());
+      products.stream().map(product -> new InsuranceStatusDTO(
+        product,
```
johantjellden: Why do we need to do this? As far as I can tell we did not do this before contracts-and-agreements.

vonElfvin: We did not do this, this is separate from C&A. It's a QoL improvement for IEX

PR #451, opened 2020-01-29
Add ability to backfill missing insurance certificates, per contract and in bulk

vonElfvin:


src/main/java/com/hedvig/contract/shared/models/repositories/AgreementStateRepository.kt
```diff
@@ -39,4 +39,15 @@ interface AgreementStateRepository<TAgreement : AgreementState> : JpaRepository<
...
     @Param("date") date: LocalDate
   ): List<AgreementState>
+
+  @Query(
+    """
+    FROM AgreementState ags
+    WHERE ags.from IS NOT NULL
+      AND (ags.certificate.key IS NULL OR ags.certificate.bucketName IS NULL)
+      AND ags.removedAt IS NULL
+      AND ags.contract.terminationDate IS NULL
```
johantjellden: Cant the termination date be in the future?

src/main/java/com/hedvig/productPricing/service/web/InsuranceCertificateController.java
```diff
@@ -103,37 +102,17 @@ public InsuranceCertificateController(
...
   }
 
   @PostMapping("/_/insurance/certificates/backfillMissing")
   public ResponseEntity<Void> backfillInsuranceCertificates() {
-    insuranceService.backfillMissingInsuranceCertificates();
-    return ResponseEntity.noContent().build();
+    throw new DeprecatedException(InsuranceCertificateController.class.getSimpleName(), "/_/insurance/certificates/backfillMissing", Collections.emptyList());
   }
 
   @PutMapping("/_/insurance/{productId}/certificates/regenerate")
```
johantjellden: Is this not deprecated as well?

vonElfvin: It's seldom used but still works and can be used in the future. Should be replaced and deprecated at some point however!

src/main/java/com/hedvig/contract/shared/models/repositories/AgreementStateRepository.kt
vonElfvin: Yes I added that 👍

src/test/java/com/hedvig/contract/testUtils/builder/ApartmentAgreementBuilder.kt
```diff
@@ -31,6 +31,19 @@ data class ApartmentAgreementBuilder(
...
+    from: LocalDate? = this.from,
+    to: LocalDate? = this.to,
+    certificate: InsuranceCertificate? = this.certificate
+  ) = copy(
+    id = id,
+    from = from,
+    to = to,
+    certificate = certificate
+  )
+
```
johantjellden: Super fun that you are using our builders 🎉 . However the idea is that `with` functions are used to pass in other entities or things that you would like to create separately in your tests.
So for these kind of changes it is fine to use the copy method directly.

src/test/java/com/hedvig/contract/testUtils/builder/ApartmentContractStateBuilder.kt
```diff
@@ -22,17 +22,16 @@ data class ApartmentContractStateBuilder(
...
   val renewals: List<RenewalState> = listOf()
 ) {
-  fun with(agreement: ApartmentAgreementStateBuilder): ApartmentContractStateBuilder {
-
-    return this.copy(agreements = setOf(agreement))
-  }
-
   fun with(
+    id: UUID = this.id,
+    terminationDate: LocalDate? = this.terminationDate,
```
johantjellden: As per the other comment these attributes should not be set in the `with` function.

PR #456, opened 2020-02-03
Contract & Agreements Reporting

vonElfvin:
- Add ability to project a contract into its underwriting years with information per date


vonElfvin:
Replaced by reporting-master

PR #463, opened 2020-02-13
Feature/build contract state from event stream for reporting

vonElfvin:


src/main/java/com/hedvig/contract/reporting/utils/EventUtils.kt
```diff
@@ -55,6 +81,156 @@ fun getOriginalTimestampOfPreviousEventIfEventBreaksOriginalTimestampChain(
...
+      contractStateFactory.createContractStateWithPendingAgreement(
+        contractId = creationEvent.contractId,
+        holderMemberId = creationEvent.holderMemberId,
+        agreement = creationEvent.agreement,
+        switchedFrom = creationEvent.switchedFrom,
+        signSource = creationEventMessage.metaData[MetadataKeys.SOURCE] as SignSource?,
+        createdAt = creationTimestamp
+      )
+    }
+    else -> throw Exception("Illegal creation event type (contractId=$contractId))")
```
palmenhq: maybe add which type it is to the message also?

src/main/java/com/hedvig/contract/web/ResetController.kt
```diff
@@ -27,4 +28,17 @@ class ResetController(
       }
     logger.info("ProcessingGroup $CONTRACT_STATE RESET")
   }
+
+  @DeleteMapping("/Reporting")
```
palmenhq: small `R` maybe? to be consistent

vonElfvin: The name is the name of the ProcessingGroup so it should be "Reporting" imo, the one above is "ContractState".

src/main/java/com/hedvig/contract/reporting/utils/EventUtils.kt
vonElfvin: 👍

src/main/java/com/hedvig/contract/shared/projections/ContractStateProjection.kt
```diff
@@ -107,24 +106,23 @@ class ContractStateProjection constructor(
 
   @EventHandler
   override fun on(
-    event: PendingAgreementActivated<BaseAgreement>,
+    event: AgreementAddedWithoutPredecessorAdjusted<BaseAgreement>,
```
vonElfvin: This file looks a bit weird since I changed the ordering to be alphabetical across the service. This one was randomly ordered.

src/main/java/com/hedvig/contract/web/ResetController.kt
palmenhq: Ahaa i see the logic. Sure 👍

vonElfvin:
> In a way i think this is way better, in another way it feels bad to propagate the event handling outside the aggregate but i think this makes sense. Whats your thoughts @meletios ?

I see what you mean. However, I would argue that there is 0 logic here, all logic present is doing what the event tells you to do.

meletios:
> > In a way i think this is way better, in another way it feels bad to propagate the event handling outside the aggregate but i think this makes sense. Whats your thoughts @meletios ?
> 
> I see what you mean. However, I would argue that there is 0 logic here, all logic present is doing what the event tells you to do.

Since there is no logic, rather than how to build the projection, i think its fine. But i totally see your point Mr Palmen.

PR #466, opened 2020-02-17
Allow parallel contracts with some minor changes

vonElfvin:


src/main/java/com/hedvig/contract/shared/models/repositories/ContractToProductRepository.kt
```diff
@@ -48,11 +52,13 @@ class ContractToProductRepository(
 
   override fun findByMemberId(memberId: String): List<ProductEntity> {
     val contracts = contractStateRepository.findByHolderMemberId(memberId)
-    val contract = if (contracts.isNotEmpty()) contracts.first() else null
+    val products = contracts.flatMap { contract ->
```
meletios: Can we rename this variable to `productsFromActiveAgreements`. Then i think it will be clear why we are doing (what we are doing) on line 60

PR #471, opened 2020-02-20
return max insurable amount and deductible

AlexSJ1990:


src/main/java/com/hedvig/contract/web/InternalContractController.kt
```diff
@@ -0,0 +1,25 @@
...
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+@RequestMapping("/_/contracts")
+class InternalContractController(
+  private val contractService: ContractService
+) {
+  @GetMapping("/{memberId}/insurableAmountInfo")
```
johantjellden: Maybe something like `/{memberId}/latest/insurableAmountInfo`

src/main/java/com/hedvig/contract/service/ContractServiceImpl.kt
```diff
@@ -69,4 +73,23 @@ class ContractServiceImpl(
       RegenerateInsuranceCertificate(contractId = contract.id, agreementId = agreementId)
     )
   }
+
+  override fun getInsurableAmount(memberId: String, category: String): InsurableAmount {
```
johantjellden: This should probably be an enum.

src/main/java/com/hedvig/contract/web/InternalContractController.kt
meletios: i would argue for` /i/members/{memberId}/contracts/latest/insurableAmountInfo`

src/main/java/com/hedvig/contract/service/ContractServiceImpl.kt
```diff
@@ -69,4 +73,23 @@ class ContractServiceImpl(
...
+          is ApartmentAgreementState -> if (agreement.lineOfBusiness.isStudent) 25_000 else 50_000
+          is HouseAgreementState -> 50_000
+          else -> throw Exception("what")
+        }
+      }
+    }
+
+    val deductible = Money.of(1500, "SEK")
+
+    return InsurableAmount(Money.of(maxAmount, "SEK"), deductible)
```
meletios: The currency should be based on the contract type i would say. So if ApartmentAgreementState then SEK if GreekApartmentInsurance then EUR

vonElfvin: I'll fix that for Norway etc.

src/main/java/com/hedvig/contract/web/InternalContractController.kt
```diff
@@ -0,0 +1,26 @@
...
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+@RequestMapping("/_/contracts")
+class InternalContractController(
+  private val contractService: ContractService
+) {
+  @GetMapping("/i/members/{memberId}/contracts/latest/insurableLimitsInfo")
```
vonElfvin:
```suggestion
  @GetMapping("/insurableLimitsInfo/member/{memberId}")
```

src/main/java/com/hedvig/contract/web/InternalContractController.kt
```diff
@@ -15,7 +15,7 @@ import org.springframework.web.bind.annotation.RestController
 class InternalContractController(
   private val contractService: ContractService
 ) {
-  @GetMapping("/i/members/{memberId}/contracts/latest/insurableLimitsInfo")
+  @GetMapping("insurable/limits/member/{memberId}")
```
meletios:
```suggestion
  @GetMapping("/members/{memberId}/insurableLimits")
```

PR #473, opened 2020-02-21
Draft schema with contracts

johantjellden:


src/main/resources/schema.graphqls
```diff
@@ -10,6 +10,228 @@ type Query {
   referralInformation: Referrals!
   """Returns redeemed campaigns belonging to authedUser"""
   redeemedCampaigns:[Campaign!]!
+
+  currentContracts: [Contract!]!
+}
+
+type Contract {
+  id: ID!
+  holderMemberId: String!
```
hedvigoscar: Are we adding this because it might differ from the current member's ID?

src/main/resources/schema.graphqls
```diff
@@ -10,6 +10,228 @@ type Query {
...
+interface AgreementCore {
+  id: ID!
+  status: AgreementStatus!
+  fromDate: LocalDate
+  toDate: LocalDate
+  premium: MonetaryAmountV2!
+  certificateUrl: String
+}
+
+enum AgreementStatus {
```
hedvigoscar: I'm not sure that I really understand the meaning of these enums.

Perhaps we could add some comments to them? This will make it appear as documentation in the schema which is really nice

src/main/resources/schema.graphqls
```diff
@@ -10,6 +10,228 @@ type Query {
...
+  INACTIVE
+  ACTIVE_IN_FUTURE
+  ACTIVE
+  ACTIVE_IN_PAST
```
sampettersson: `ACTIVE_IN_PAST` and `TERMINATED` seem mutually exclusive or? i.e it's not really clear what they do.

src/main/resources/schema.graphqls
```diff
@@ -10,6 +10,228 @@ type Query {
...
+  certificateUrl: String
+  status: AgreementStatus!
+  address: Address
+  numberCoInsured: Int!
+  squareMeters: Int!
+  type: NorwegianHomeContentLineOfBusiness
+}
+
+enum NorwegianHomeContentLineOfBusiness {
+  STUDENT_RENT
```
hedvigoscar: Are these accurate lines of business for Norway? 🤔

src/main/resources/schema.graphqls
sampettersson: can this be different than the member we are logged in as?

src/main/resources/schema.graphqls
```diff
@@ -100,3 +322,5 @@ type MonetaryAmountV2 {
 
 # A String-representation of `java.time.LocalDate`, ex:  `"2018-09-26"`
 scalar LocalDate
+# A Unix timestamp-representation of `java.time.Instant`. ex: `"1582894553"`
```
hedvigoscar: 😍

src/main/resources/schema.graphqls
```diff
@@ -10,6 +10,228 @@ type Query {
...
+
+  currentContracts: [Contract!]!
+}
+
+type Contract {
+  id: ID!
+  holderMemberId: String!
+  status: ContractStatus
+  currentAgreement: Agreement!
+  masterInception: LocalDate
```
sampettersson: it's unclear what this means and does.

src/main/resources/schema.graphqls
```diff
@@ -10,6 +10,228 @@ type Query {
...
+  terminationDate: LocalDate
```
sampettersson: sometimes we suffix with date, and sometimes we don't, either is fine, but we should be consistent.

sampettersson: I prefer not adding `date` tho as it's already clear from the type.

src/main/resources/schema.graphqls
```diff
@@ -10,6 +10,228 @@ type Query {
   referralInformation: Referrals!
   """Returns redeemed campaigns belonging to authedUser"""
   redeemedCampaigns:[Campaign!]!
+
+  currentContracts: [Contract!]!
```
sampettersson: just `contracts`? as this could contain `terminated` ones as well right?

src/main/resources/schema.graphqls
```diff
@@ -10,6 +10,228 @@ type Query {
...
+  createdAt: Instant
+}
+
+enum ContractStatus {
+  PENDING,
+  ACTIVE,
+  TERMINATED
+}
+
+type Renewal {
```
sampettersson: We already have this type in the schema, either you need to make this `RenewalV2` or we need to have the exact same fields, else it's gonna be a mess hehe

src/main/resources/schema.graphqls
```diff
@@ -10,6 +10,228 @@ type Query {
...
+  renewalDate: LocalDate
+  draftCertificateUrl: String
+}
+
+interface AgreementCore {
+  id: ID!
+  status: AgreementStatus!
+  fromDate: LocalDate
```
sampettersson: `activeFrom`?

src/main/resources/schema.graphqls
```diff
@@ -10,6 +10,228 @@ type Query {
...
+  toDate: LocalDate
```
sampettersson: `activeTo`?

src/main/resources/schema.graphqls
```diff
@@ -10,6 +10,228 @@ type Query {
...
+  premium: MonetaryAmountV2!
```
sampettersson: we use the term `cost` in other places in the schema, should we do that here too?

src/main/resources/schema.graphqls
vonElfvin: With Co-Insured it could very much change in the future. However, it might not be necessary to be included for now.

src/main/resources/schema.graphqls
```diff
@@ -10,6 +10,228 @@ type Query {
...
+  address: Address
+  numberCoInsured: Int!
+  squareMeters: Int!
+  type: NorwegianHomeContentLineOfBusiness
+}
+
+enum NorwegianHomeContentLineOfBusiness {
+  STUDENT_RENT
+  RENT
+  STUDENT_OWN
```
sampettersson: For sweden we use the term `BRF` for this, should we maybe use the actual norwegian term too?

src/main/resources/schema.graphqls
hedvigoscar: I'm not questioning if it should be there, I was just not entirely sure about the purpose 🙂

I think it's a good idea to keep it actually

src/main/resources/schema.graphqls
vonElfvin: For me, in the the backend at least, the "cost" is what you actually pay a given month. For example 100 SEK premium for half a month is ~50 SEK cost. The premium never changes but the cost may.

src/main/resources/schema.graphqls
vonElfvin: This is an attempt to introduce a new "concept" where we now see a contract as terminated or not. For the app's case the terminated ones are terminated insurances. And "active_in_past" is an agreement that was active before but the insurance is still active with another agreement today.

vonElfvin: The risk of having TERMINATED=ACTIVE_IN_PAST is that we do not distinguish these two cases.

vonElfvin: From the app's perspective it could be argued that an agreement is either INACTIVE, ACTIVE_IN_FUTURE, ACTIVE or TERMINATED, since if and agreement is "ACTIVE_IN_PAST" then we would return the future agreement. I kept it here for future cases where we want to return all agreements. What do you think?

src/main/resources/schema.graphqls
vonElfvin: Yes

src/main/resources/schema.graphqls
vonElfvin: I personally think we should call the Swedish one OWN but we are stuck with BRF now. But I'm not the one that decided these types regardless, they're used in underwriter already as well.

src/main/resources/schema.graphqls
vonElfvin: That's the old term yes, now we have a from date and a to date. Would you prefer us sending the old term or the new?

src/main/resources/schema.graphqls
vonElfvin: 👍

src/main/resources/schema.graphqls
vonElfvin: In the future with co-insured yes, not now no. Should I remove it you think?

src/main/resources/schema.graphqls
vonElfvin: It's an insurance term, it's the first date your insurance is/was active from.

src/main/resources/schema.graphqls
hedvigoscar: Ok!

I thought that the Innbo-insurance product was the same for rentals and owned apartments? 🤔

src/main/resources/schema.graphqls
vonElfvin: The only risk I see with not calling it `date` is if we introduce a type called Termination, similar to renewalDate where a renewal contains a certificateUrl but also a date for the renewal.

src/main/resources/schema.graphqls
sampettersson: hmm, no then maybe we shall keep it, but rename it to `holderMember: ID!` maybe?

src/main/resources/schema.graphqls
vonElfvin: That's travel (as far as I know)

src/main/resources/schema.graphqls
vonElfvin: Smart!

src/main/resources/schema.graphqls
vonElfvin: Will look into it

src/main/resources/schema.graphqls
hedvigoscar: I was under the impression that you in Norway get your 'bostadsrättstillägg' by your BRF-equivalent and hence Innbo is the same for rentals and owned apartments. 

I could of course be wrong. Let's make sure we know what's right here before we merge 🙂

src/main/resources/schema.graphqls
sampettersson: Ah, then it makes sense! Maybe you can use naming to make it a bit clearer, like `RENEWED` or `REPLACED` or something similar?

src/main/resources/schema.graphqls
sampettersson: I think that maybe the old term made more sense, because what does `fromDate` really mean? i.e what is `from`, (active, decided, created, bought)?. `activeFrom: LocalDate` both tells me "it's a date" + "it was active from that date".

src/main/resources/schema.graphqls
sampettersson: In Sweden we actually have two types of owned apartments, Bostadsrätt and Äganderätt. Are we sure that is not the case in Norway? That is why I think it makes the most sense to name it for what actual type it is.

src/main/resources/schema.graphqls
vonElfvin: Maybe @FredrikAreschoug or @meletios can chip in here? I don't know tbh

src/main/resources/schema.graphqls
sampettersson: ☝️I guess the above is on you @FredrikAreschoug as you are doing underwriter? 😄

src/main/resources/schema.graphqls
sampettersson: makes sense.

src/main/resources/schema.graphqls
vonElfvin: An idea is to keep it as ACTIVE_IN_PAST but remove it from the enum sent to the app, and rather send "ACTIVE" if you ever receive one (which you never should)

src/main/resources/schema.graphqls
FredrikAreschoug: True the question here is if it affects the contract in any way. If so let's change to brf. I the quote we should probably change it or remove it. I haven't got an answer on if it's going to affect the price or not

FredrikAreschoug: If `Ägarbostad`, `Andelslägenhet/bostadsrätt` and `aktielägenhet` are all the same to us in the contract and pricing I see no reason to split this up.

FredrikAreschoug: Talked to Tomas about this and we do not care about how thay own the apartment. But we do care about if they own or rent

src/main/resources/schema.graphqls
sampettersson: I think removing it is best then :)

src/main/resources/schema.graphqls
```diff
@@ -10,6 +10,240 @@ type Query {
...
+  id: ID!
+  activeFrom: LocalDate
+  activeTo: LocalDate
+  premium: MonetaryAmountV2!
+  certificateUrl: String
+  status: AgreementStatus!
+  numberCoInsured: Int!
+}
+
+union ExtraBuilding =
```
sampettersson: should we scrap this and just have a type which is simply `ExtraBuilding`? we don't currently differentiate between different types of buildings anywhere AFAIK, and this quite messy 😬 what you think @hedvigoscar ?

hedvigoscar: Right now there seems to be no particular reason to have them, but I do remember us having some slight reason for including them when designing the schema. @FredrikAreschoug Do you know anything more? 🙂

FredrikAreschoug: A type that is called just ExtraBuilding will probably break the stitching. If you don't want the information you will have to create a V2 of it :)

vonElfvin: With this I conclude that I should keep it as it is, or do you agree?

src/main/resources/schema.graphqls
vonElfvin: Changed it to "UpcomingRenewal", is that a good name you think?

src/main/resources/schema.graphqls
vonElfvin: I changned it to termination

src/main/resources/schema.graphqls
vonElfvin: Changed it to "inception" now

src/main/resources/schema.graphqls
vonElfvin: Will do

src/main/resources/schema.graphqls
FredrikAreschoug: @sampettersson @hedvigoscar you are the consumers as is or `ExtraBuildingV2`?

PR #480, opened 2020-03-05
Feature/norwegian contracts

vonElfvin:
- Wrestle a bit with ExtraBuildingDto
- Include agreementId when creating a contract to be stored in underwriter
- Add tests to `AgreementStatus`


src/main/java/com/hedvig/contract/shared/models/objects/ExtraBuilding.kt
```diff
@@ -45,6 +45,14 @@ class ExtraBuilding(
     ExtraBuildingType.OTHER -> "Övrigt"
   }
 
+  fun toExtraBuildingDto(localizationService: LocalizationService? = null, locale: Locale? = null) = ExtraBuildingDto(
```
FredrikAreschoug: Feels strange to default `localizationService` and `locale` to null we should be more explicit when we don't want to translate

FredrikAreschoug: maybe creating a new function called toExtraBuildingEmptyDisplayNameDto when we really don't care about translations

src/main/java/com/hedvig/contract/web/dtos/AgreementQuote.kt
```diff
@@ -18,59 +17,59 @@ import java.util.UUID
...
-  val currency: String
-) {
-  class SwedishApartmentQuote(
-    quoteId: UUID,
-    fromDate: LocalDate?,
-    toDate: LocalDate?,
-    premium: BigDecimal,
-    currency: String,
+sealed class AgreementQuote {
+  abstract val quoteId: UUID
```
FredrikAreschoug: :heart_eyes:

src/main/java/com/hedvig/contract/shared/models/objects/ExtraBuilding.kt
vonElfvin: Split them up now

PR #481, opened 2020-03-06
Add Query for InsuranceCost, including when having multiple contracts

vonElfvin:


src/main/java/com/hedvig/contract/service/ContractServiceImpl.kt
```diff
@@ -58,6 +59,19 @@ class ContractServiceImpl(
     return contractStateRepository.findByHolderMemberId(holderMemberId)
   }
 
+  override fun getTotalCostOfNonTerminatedContractsByHolderMemberId(holderMemberId: String): MonetaryAmount? {
+    val byCurrency = getContractStatesByHolderMemberId(holderMemberId)
```
vonElfvin: Better naming

src/main/resources/schema.graphqls
```diff
@@ -12,6 +12,8 @@ type Query {
   redeemedCampaigns:[Campaign!]!
   """Returns all contracts the member currently holds, regardless of activation/termination status"""
   contracts: [Contract!]!
+  """Returns the aggregated insurance cost of a member's PENDING, ACTIVE or ACTIVE_IN_FUTURE current agreements"""
```
meletios:
```suggestion
  """Returns the aggregated insurance cost of a member's PENDING, ACTIVE or ACTIVE_IN_FUTURE agreements"""
```

vonElfvin: It's actually supposed to be only the "current" ones, not the sum of all of them.

PR #482, opened 2020-03-09
Update dtos sent to back-office/graphql and add GraphQL queries

vonElfvin:


src/main/resources/schema.graphqls
```diff
@@ -14,12 +14,17 @@ type Query {
   contracts: [Contract!]!
   """Returns the aggregated insurance cost of a member's PENDING, ACTIVE or ACTIVE_IN_FUTURE current agreements"""
   insuranceCost: InsuranceCost
+  """Returns wether a member has at least one contract"""
```
meletios:
```suggestion
  """Returns whether a member has at least one contract"""
```

src/main/resources/schema.graphqls
```diff
@@ -14,12 +14,17 @@ type Query {
...
+  hasContract: Boolean!
+  """Returns wether a member is eligible to create a claim, i.e. if a member has an active contract"""
```
meletios:
```suggestion
  """Returns whether a member is eligible to create a claim, i.e. if a member has an active contract"""
```

vonElfvin: I'll fix it in my current update and include it in a future PR 🛩

PR #488, opened 2020-03-17
Bundle insurance cost endpoint

FredrikAreschoug:
Stuff we probably need to add:
- getActiveDiscounts
- getBillingDiscount


FredrikAreschoug:
hmm the campaign tests are failing locally :/

PR #503, opened 2020-04-01
Feature/include bundle discount on active

FredrikAreschoug:


src/main/java/com/hedvig/contract/service/PremiumServiceImpl.kt
```diff
@@ -28,6 +32,15 @@ class PremiumServiceImpl(
     }
   }
 
+  override fun calculateTotalPremiumCostsOfPeriod(memberId: String, period: YearMonth): MonetaryAmount {
+    val agreementPremiumCosts = calculatePremiumCostsOfPeriod(memberId, period)
+    return if (agreementPremiumCosts.isEmpty()) {
+      Money.of(BigDecimal.ZERO, "SEK")
```
meletios, line 38: 🤔 is it going to be sad with only _**SEK**_

src/main/java/com/hedvig/contract/shared/models/ContractState.kt
```diff
@@ -101,6 +103,25 @@ abstract class ContractState<TAgreement : AgreementState>(
       }
     }
 
+  val currentOrAgreementStartingThisMonth: TAgreement?
+    get() {
+      val today = LocalDate.now(DateUtil.STOCKHOLM_ZONE_ID)
```
meletios, line 108: We can get the timezone from the contract, just to be 100% legit. No need to change that though

src/main/java/com/hedvig/productPricing/campaign/service/CampaignServiceImpl.kt
```diff
@@ -86,13 +99,20 @@ class CampaignServiceImpl(
...
+  }
+
+  private fun getBundleDiscountOrNull(typesToBeBundled: List<InsuranceType>, currency: CurrencyUnit): Campaign? = if (
+    typesToBeBundled.size == 2
+    && typesToBeBundled.any { it == InsuranceType.NORWEGIAN_YOUTH_HOME_CONTENTS }
+    && typesToBeBundled.any { it == InsuranceType.NORWEGIAN_YOUTH_TRAVEL }
     && currency.currencyCode == "NOK"
   ) {
-    Money.of(BigDecimal(49), currency)
+    Campaign(_code = "NorwegianYouthBundle", incentive = Incentive.MonthlyCostDeduction(MonetaryAmountV2(BigDecimal(49).toString(), currency.currencyCode)))
```
meletios, line 115: Should we add the actual amount into a config?

src/main/java/com/hedvig/productPricing/campaign/service/CampaignServiceImpl.kt
```diff
@@ -175,13 +209,17 @@ class CampaignServiceImpl(
...
+      val premiumTotalCosts = premiumService.calculateTotalPremiumCostsOfPeriod(memberId, YearMonth.now())
+      currentTotalPrice = premiumTotalCosts.asBigDecimal()
+      preferredCurrency = premiumTotalCosts.currency.currencyCode
 
     } else {
+      /*
+        TODO: we need to figure out how not use /latestQuote, or `getActiveDiscounts` should not be used used when
+        there is no contracts let's try to remove this at some point and redirect them to `getInsuranceCostOnPrice` or
+        `getBundledInsuranceCostOnPrice`
+      */
```
meletios, line 222: 🤔

meletios: is it comments for the future? Should we log something if we end up there?

src/main/java/com/hedvig/productPricing/campaign/service/CampaignServiceImpl.kt
FredrikAreschoug: yeah why not

src/main/java/com/hedvig/productPricing/campaign/service/CampaignServiceImpl.kt
FredrikAreschoug: Yes we could log the stack trace to see if it used

FredrikAreschoug:
Refactored some in f908513 so new reviews are needed :)

PR #504, opened 2020-04-01
add endpoint to get type of contract

AlexSJ1990:


src/main/java/com/hedvig/contract/web/InternalContractController.kt
```diff
@@ -113,4 +113,14 @@ class InternalContractController(
       ResponseEntity.notFound().build()
     }
   }
+
+  @GetMapping("/members/{memberId}/contract/contractTypes")
+  fun getContractTypesForMember(@PathVariable memberId: String): ResponseEntity<List<TypeOfContract>> {
+    val contractTypes = contractService.getTypeOfContractByHolderMemberId(memberId)
+    return if (contractTypes != null) {
```
vonElfvin, line 120: Can they even be null? Maybe you changed it to throw a RuntimeException instead? Since it returns `List<TypeOfContract>`

src/main/java/com/hedvig/contract/shared/models/ContractState.kt
```diff
@@ -127,6 +134,8 @@ abstract class ContractState<TAgreement : AgreementState>(
     findAgreementOrThrow(agreementId).removedAt = removedAt
   }
 
+  abstract fun getTypeOfContract(): TypeOfContract
```
vonElfvin, line 137: 😍

PR #505, opened 2020-04-01
I think this does the trick

FredrikAreschoug:


src/main/java/com/hedvig/productPricing/campaign/service/CampaignBillingServiceImpl.kt
```diff
@@ -71,7 +81,27 @@ class CampaignBillingServiceImpl(
   override fun getBillingDiscount(memberId: String, month: YearMonth): List<CampaignDiscount> {
     val listOfCampaignDiscount = mutableListOf<CampaignDiscount>()
 
-    val data = productPricingGateway.getBillableProductSubscriptions(memberId, month)
+    val data = premiumService.calculatePremiumCostsOfPeriod(memberId, month).mapNotNull {
+      val contract = contractService.getContractStateByAgreementId(it.agreementId)
```
johantjellden, line 85: Maybe we should encapsulat this logic somewhere else. To make it easier to test and a bit clearer of what it does.

src/main/java/com/hedvig/productPricing/service/service/NewInsuranceBillingServiceImpl.kt
```diff
@@ -0,0 +1,45 @@
...
+class NewInsuranceBillingServiceImpl(
+  val premiumService: PremiumService,
+  val contractService: ContractService
+) : NewInsuranceBillingService {
+
+  override fun getMonthlyProductSubscriptionsByMemberId(memberId: String, period: YearMonth): List<BillableProductSubscription> {
+    return premiumService.calculatePremiumCostsOfPeriod(memberId, period).map {
+      val contract = contractService.getContractStateByAgreementId(it.agreementId)
+
+      val agreement = contract.pendingOrCurrentOrNextOrLastAgreement
```
FredrikAreschoug, line 25: @vonElfvin @meletios This Is the only thing I'm still unsure about can I use `pendingOrCurrentOrNextOrLastAgreement` here should I instead use something that is more like `currentOrNextOrLastOrPendingAgreement` or `currentOrNextOrLastAgreement`

meletios: Yeah since pendingAgreement dont have `from` and `to` dates , it will explode on line 29. So you need only agreements for an active contract.

meletios: But you have the agreementId available, so you can just fetch the agreement based on the id, right?

src/main/java/com/hedvig/contract/customerio/Extensions.kt
```diff
@@ -1,9 +1,15 @@
...
 import java.time.ZoneId
+import javax.money.MonetaryAmount
 
 fun LocalDate.toStockholmInstant() : Instant {
   return this.atStartOfDay(ZoneId.of("Europe/Stockholm")).toInstant()
 }
+
+fun MonetaryAmount.asBigDecimal(): BigDecimal {
+  return this.number.numberValue(BigDecimal::class.java)
+}
```
vonElfvin, line 15: A bit weird to have this in the `customerio/Extensions.kt` file

src/main/java/com/hedvig/productPricing/service/service/NewInsuranceBillingServiceImpl.kt
```diff
@@ -0,0 +1,45 @@
...
+import com.hedvig.contract.type.swedishApartment.models.SwedishApartmentContractState
+import com.hedvig.contract.type.swedishHouse.models.SwedishHouseContractState
+import com.hedvig.productPricing.campaign.integration.productPricing.BillableProductSubscription
+import org.springframework.stereotype.Service
+import java.lang.RuntimeException
+import java.time.YearMonth
+import java.time.ZoneId
+
+@Service
+class NewInsuranceBillingServiceImpl(
```
meletios: ContractBillingService better? 

its a bit confusing the `NewInsuranceBillingServiceImpl` right?

src/main/java/com/hedvig/productPricing/campaign/integration/productPricing/PremiumBillingMapperImpl.kt
```diff
@@ -0,0 +1,43 @@
...
+import java.time.YearMonth
+import java.time.ZoneId
+
+@Service
+class PremiumBillingMapperImpl(
+  val premiumService: PremiumService,
+  val contractService: ContractService
+) : PremiumBillingMapper {
+
+  override fun getMonthlyBillableSubscriptionsByMemberId(memberId: String, period: YearMonth): List<BillableSubscription> {
```
vonElfvin, line 20: I like having it in campaign-service over product-pricing, however I think this would fit nicely inside PremiumService. I.e. go straight from `CampaignBillingServiceImpl` to `PremiumService` and skip this mapper entirely

src/main/java/com/hedvig/productPricing/campaign/domain/model/Member.kt
```diff
@@ -240,11 +240,11 @@ open class Member(
...
-      month: YearMonth,
-      billablePeriods: List<BillableProductSubscription>,
-      percentageDiscount: BigDecimal,
-      redeemedCampaignActiveAt: LocalDate,
-      redeemedCampaignEndDate: LocalDate?): MonetaryAmount? {
+        month: YearMonth,
+        billablePeriods: List<BillableSubscription>,
+        percentageDiscount: BigDecimal,
+        redeemedCampaignActiveAt: LocalDate,
+        redeemedCampaignEndDate: LocalDate?): MonetaryAmount? {
```
meletios, line 247: White spaces

src/main/java/com/hedvig/productPricing/service/service/NewInsuranceBillingServiceImpl.kt
meletios: this was from before, please discard it

src/main/java/com/hedvig/productPricing/campaign/integration/productPricing/PremiumBillingMapperImpl.kt
FredrikAreschoug: haha then you need to battle @johantjellden ```Maybe we should encapsulat this logic somewhere else. To make it easier to test and a bit clearer of what it does.```

vonElfvin: What I'm saying is that we should encapsulate it but do it in `ContractService` and thus not disagreeing with Johan.

vonElfvin: Now we are spreading out the logic instead of keeping the premium calculations to PremiumService.

PR #506, opened 2020-04-02
return empty list for TypesOfContract if member doesn't have a contract

AlexSJ1990:


src/main/java/com/hedvig/contract/service/ContractServiceImpl.kt
```diff
@@ -275,7 +275,7 @@ class ContractServiceImpl(
     val contractStates = getContractStatesByHolderMemberId(holderMemberId)
 
     if(contractStates.isEmpty()) {
-      throw java.lang.RuntimeException("No contracts exist for member $holderMemberId")
+      return emptyList()
     }
     return contractStates.map{ contractState -> contractState.getTypeOfContract() }
```
vonElfvin, line 280: Since the map will return an empty list if the list is empty lines 277 to 279 are redundant

PR #512, opened 2020-04-08
hasContract endpoint

FredrikAreschoug:


src/main/java/com/hedvig/productPricing/service/serviceIntegration/memberService/MemberServiceImpl.java
```diff
@@ -43,4 +44,13 @@ public String getPickedLocale(String memberId) {
       throw e;
     }
   }
+
+  @Override
+  public void notifyContractsCreated(String memberId) {
```
meletios, line 49:
```suggestion
  public void notifyContractCreated(String memberId) {
```

Its only one per time

src/main/java/com/hedvig/productPricing/service/serviceIntegration/memberService/MemberServiceImpl.java
```diff
@@ -43,4 +44,13 @@ public String getPickedLocale(String memberId) {
...
+    try {
+      memberServiceClient.contractsCreated(memberId);
+    } catch (FeignException e) {
```
meletios, line 52:
```suggestion
    } catch (Exception e) {
```

feels safer

PR #520, opened 2020-04-16
percentage discount on norwegain bundle

FredrikAreschoug:


src/main/java/com/hedvig/productPricing/campaign/service/BundleDiscountsServiceImpl.kt
```diff
@@ -58,25 +63,46 @@ class BundleDiscountsServiceImpl(
...
+        val amount = bundleDiscount.amount.multiply(BigDecimal(activePeriodInPercentage)).setScale(0, RoundingMode.HALF_EVEN)
+        if (amount > BigDecimal.ZERO) {
+          BundleDiscount.MonthlyCostDeduction(
+            bundleDiscount.identifier,
+            amount
+          )
+        } else null
+      }
+      is BundleDiscount.PercentageDiscount -> {
+        var isAllContractsActive = true
```
meletios, line 80:
```suggestion
        var areAllContractsActive = true
```

meletios: 😇

src/main/java/com/hedvig/productPricing/campaign/service/BundleDiscountsServiceImpl.kt
```diff
@@ -58,25 +63,46 @@ class BundleDiscountsServiceImpl(
...
+        contracts.forEach { contract ->
+          if (!agreementActiveDuringMonth(contract.pendingOrCurrentOrNextOrLastAgreement, yearMonth)) {
+            isAllContractsActive = false
+            return@forEach
+          }
+        }
```
meletios, line 86: shall we use any or none instead of a foreach? Just as improvement

PR #521, opened 2020-04-17
Add norwegian home content youth LOB

vonElfvin:


src/main/java/com/hedvig/contract/reporting/models/ReportingLineOfBusiness.kt
```diff
@@ -5,5 +5,6 @@ enum class ReportingLineOfBusiness {
   HOUSEHOLD_STUDENT,
   HOUSEHOLD_HOUSE,
   NORWEGIAN_HOME_CONTENT,
+  NORWEGIAN_HOME_CONTENT_YOUTH,
   NORWEGIAN_TRAVEL
```
akashpatl, line 9: Let's add TRAVEL_YOUTH

PR #523, opened 2020-04-23
Refactored some to deprecate and remove top most logger

FredrikAreschoug:
removes logging of this -> https://sentry.io/organizations/hedvig/issues/1614464751/?query=is%3Aunresolved

And depricates all of the path to it. Let me know what you think :)


src/main/java/com/hedvig/productPricing/campaign/service/CampaignServiceImpl.kt
```diff
@@ -194,7 +184,8 @@ class CampaignServiceImpl(
     return redeemedCampaignRepository.countActiveCampaignsPerMemberId(memberId)
   }
 
-  private fun getCurrentInsuranceCost(memberId: String): Money {
+  @Deprecated("Currently only used by depricated getActiveDiscounts")
```
AlexSJ1990, line 187: small typo should be deprecated

PR #531, opened 2020-04-30
return agreement based on id rather than current pending or last

AlexSJ1990:


src/main/java/com/hedvig/contract/service/AgreementServiceImpl.kt
```diff
@@ -119,6 +119,8 @@ class AgreementServiceImpl(
   }
 
   override fun getAgreementFromAgreementId(agreementId: UUID): AgreementState {
-    return contractService.getContractStateByAgreementId(agreementId).pendingOrCurrentOrNextOrLastAgreement
+    return contractService.getContractStateByAgreementId(agreementId).agreementsTimeline.find {
+      agreement -> agreement.id == agreementId
+    }!!
```
vonElfvin, line 124:
```suggestion
    return agreementRepository.findByIdOrNull(agreementId)
      ?: throw AgreementNotFoundException(...)
```

src/main/java/com/hedvig/contract/service/AgreementServiceImpl.kt
```diff
@@ -119,6 +123,7 @@ class AgreementServiceImpl(
   }
 
   override fun getAgreementFromAgreementId(agreementId: UUID): AgreementState {
-    return contractService.getContractStateByAgreementId(agreementId).pendingOrCurrentOrNextOrLastAgreement
+    return agreementStateRepository.findByIdOrNull(agreementId)
+      ?: throw AgreementNotFoundException("Unable to find agreement with agreement id: $agreementId")
```
vonElfvin, line 127: We should probably update the `AgreementNotFoundException` to only take the `agreementId` and write out that message for you. This is fine for now though.

PR #534, opened 2020-05-06
Feature/notification service events

johantjellden:


src/main/java/com/hedvig/contract/customerio/EventListenerStrategy.kt
```diff
@@ -0,0 +1,174 @@
...
+
+  override fun onContractWithPendingAgreementCreated() {
+    this.execWithCurrentOrLastContract(contract, sequenceNumber)
+  }
+
+  override fun onAgreementAddedWithoutPredecessorAdjusted(
+    contract: ContractState<AgreementState>,
+    sequenceNumber: Long
+  ) {
+    //NOOP
```
vonElfvin, line 163:
```suggestion
    // noop
```

src/main/java/com/hedvig/contract/customerio/EventListenerStrategy.kt
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  override fun onAgreementAddedWithPredecessorAdjusted(contract: ContractState<AgreementState>, sequenceNumber: Long) {
+    //NOOP
+  }
+
+  override fun onAgreementReplaced(contract: ContractState<AgreementState>, sequenceNumber: Long) {
+    notificationService.sendStartDateUpdated(contract.id, contract.masterInception, sequenceNumber)
+  }
```
vonElfvin, line 172: There is no way the startdate would change when we replace an agreement?

src/main/java/com/hedvig/contract/customerio/EventListenerStrategy.kt
```diff
@@ -0,0 +1,174 @@
...
+class Norway(
+  private val notificationService: NotificationService,
+  private val contract: ContractState<AgreementState>,
+  private val sequenceNumber: Long?
+) : EventListenerStrategy {
+  private fun execWithCurrentOrLastContract(
+    contract: ContractState<AgreementState>,
+    sequenceNumber: Long?
+  ) {
+    val possibleMasterInceptionDate = if (!contract.hasPendingAgreement) contract.masterInception else null
```
vonElfvin, line 127: There is a more explicit function called
```suggestion
    val possibleMasterInceptionDate = if (contract.hasMasterInception) contract.masterInception else null
```

src/main/java/com/hedvig/contract/shared/events/PendingAgreementActivated.kt
```diff
@@ -1,9 +1,9 @@
 package com.hedvig.contract.shared.events
 
-import com.hedvig.contract.shared.aggregates.BaseAgreement
+import com.hedvig.contract.shared.stereotypes.AgreementStereotype
 import java.util.UUID
 
-class PendingAgreementActivated<TAgreement : BaseAgreement>(
+class PendingAgreementActivated<TAgreement : AgreementStereotype>(
```
vonElfvin, line 6: This should not be changed. This way there could in theory exist an event with SwedishApartmentAgreementState which should not be allowed.

```suggestion
class PendingAgreementActivated<TAgreement : BaseAgreement>(
```

src/main/java/com/hedvig/contract/customerio/EventListenerStrategy.kt
vonElfvin: Very important to me

PR #538, opened 2020-05-08
Feature/bundle discount

vonElfvin:
This is the logic on how we will track the bundle discount inside the Contract. Around 1200 of the lines of from tests.

The logic is pretty easy but the code a bit hard to read:
- It will start a bundle discount on applicable agreements in the correct contract on events
- It will unbundle the other contract if one is terminated. It will then remove the bundle discount from its future agreements 
- If the bundle discount is start/stopped in the middle of an agreement period it will "split it" to two identicial ones and add the bundle discount on the latter
- All bundle related events are applied independently from the agreement events, also the agreement events are ensured to not include any bundle discount information
- The logic of if a contract can be bundled or not is owned by `ContractBundleService`
- The logic to listen for contract events to potentially bundle/unbundle two contracts is owned by `ContractBundleSaga`
- The logic to calculate the price for the agreement based on it's type is owned by `BundleDiscountService`
- I moved the "getContractStateFromTimestamp" to a class `ContractStateStream` that is able to build a contract state from the `EventStore` instead of reading in the database. This guarantees that the contract is available and goes around the eventual consistency issue when handing events. If this works well we should explore doing this more in the future. It has done wonders in reporting already.


FredrikAreschoug:
This looks good and all and I don't want to sound like a broken record but I really don't see the benefit of having premium discounts inside of agreements. I would rather see it as a ledger where you have agreements and whenever we add new agreement(or whatever else happens that triggers adding a new premium discount) that member gets a premium discount. It feels like a way more flexible solution and I feel like it will reduce the complexity and not having to take care of your third point. But yeah let me know if you feel that I'm in the wrong here and I will agree and commit!

vonElfvin:
> This looks good and all and I don't want to sound like a broken record but I really don't see the benefit of having premium discounts inside of agreements. I would rather see it as a ledger where you have agreements and whenever we add new agreement(or whatever else happens that triggers adding a new premium discount) that member gets a premium discount. It feels like a way more flexible solution and I feel like it will reduce the complexity and not having to take care of your third point. But yeah let me know if you feel that I'm in the wrong here and I will agree and commit!

I see what you mean but I did actually try to have the separately but the complexity increased an insane amount that is quite tricky to explain over a message like this. I would like to say that i ONLY have to do the third point makes this solution extremely powerful.

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
```diff
@@ -0,0 +1,111 @@
...
+    val firstContract = contracts.first()
+    val secondContract = contracts.last()
+
+    val sameMember = firstContract.holderMemberId == secondContract.holderMemberId
+    if (!sameMember) {
+      return false
+    }
+
+    val validTypes = when {
+      firstContract is NorwegianHomeContentContractState && secondContract is NorwegianTravelContractState -> true
```
akashpatl, line 46: `contracts.contains(NorwegianHomeContentContractState) && contracts.contains(NorwegianTravelContractState) `could replace these mirror checks

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
```diff
@@ -0,0 +1,111 @@
...
+    if (!canBundle(contracts)) {
+      return
+    }
+    bundleContracts(contracts, commandGateway, event, eventTimestamp)
+  }
+
+  fun canBundle(
+    contracts: List<ContractState<AgreementState>>
+  ): Boolean {
+    if (contracts.size != 2) {
```
akashpatl, line 34: In the future we will have members with more than 2 contracts and two of them should be bundled. I.e if someone had home insurance, travel and object insurance. Something to consider but perhaps this is good enough for now.

src/main/java/com/hedvig/contract/shared/models/objects/TerminationReason.kt
```diff
@@ -18,6 +18,7 @@ enum class TerminationReason {
   MOVED,
   MOVED_ABROAD,
   MOVED_IN_WITH_PARENTS,
+  MOVED_IN_WITH_PARTNER,
```
akashpatl, line 21: Didn't see this one earlier but there already exists a `COVERED_BY_PARTNERS_INSURANCE`. Maybe confusing to have both?

src/main/java/com/hedvig/contract/shared/models/objects/TerminationReason.kt
```diff
@@ -18,6 +18,7 @@ enum class TerminationReason {
   MOVED,
```
akashpatl, line 18: If you're already in here, could you add a termination reason like `GOT_BETTER_OFFER_FROM_ANOTHER_INSURER` or something similar?

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
vonElfvin: Yea I know my idea was this way we can at least change it in the future

src/main/java/com/hedvig/contract/shared/models/objects/TerminationReason.kt
vonElfvin: For some reason that is too confusing?

src/main/java/com/hedvig/contract/shared/models/objects/TerminationReason.kt
```diff
@@ -26,5 +27,6 @@ enum class TerminationReason {
   CONFIRMED_FRAUD,
   SUSPECTED_FRAUD,
   OTHER,
+  TEST,
```
meletios, line 30: AWESOME

src/main/java/com/hedvig/contract/shared/models/objects/TerminationReason.kt
vonElfvin: I would say those are kind of two different things though. One of them is moving out and the other is them already living with their partner. At least this way we will give IEX a good option to pick when they are in the scenario and we can bundle them for future analysis.

src/main/java/com/hedvig/contract/shared/models/objects/TerminationReason.kt
vonElfvin: Added it

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
vonElfvin: The tricky thing going forward is that there is not guaranteed to be a list with both contracts. The eventual consistency fucks it up a bit :( Please check the updated solution

src/main/java/com/hedvig/contract/shared/streams/ContractStateStream.kt
```diff
@@ -0,0 +1,194 @@
...
+import com.hedvig.contract.util.MetadataKeys
+import com.hedvig.contract.util.getEventMessagesUntil
+import com.hedvig.contract.util.isOfEvent
+import org.axonframework.eventsourcing.eventstore.EventStore
+import org.springframework.stereotype.Component
+import java.time.Instant
+import java.util.UUID
+
+@Component
+class ContractStateStream(
```
vonElfvin, line 42: This file is just moved from `EventUtils` to a separate component. The logic here is already battle tested.

src/main/java/com/hedvig/contract/discounts/aggregates/BundlePremiumDiscount.kt
```diff
@@ -0,0 +1,23 @@
...
+import com.hedvig.contract.discounts.stereotypes.BundlePremiumDiscountStereotype
+import java.math.BigDecimal
+import java.util.UUID
+import javax.money.MonetaryAmount
+
+data class BundlePremiumDiscount(
+  override val bundledWithContractId: UUID,
+  override val type: BundlePremiumDiscountType,
+  override val value: BigDecimal,
+  override val amount: MonetaryAmount
```
meletios, line 14:
```suggestion
  override val percentageValue: BigDecimal,
  override val monetaryValue: MonetaryAmount
```

src/main/java/com/hedvig/contract/graphql/resolvers/ContractQuery.kt
```diff
@@ -23,7 +23,7 @@ import java.time.YearMonth
 @Transactional(readOnly = true)
 class ContractQuery(
   private val contractService: ContractService,
-  private val bundleDiscountsService: BundleDiscountsService,
+  private val bundleDiscountsService: BundleDiscountService,
```
meletios, line 26:
```suggestion
  private val bundleDiscountService: BundleDiscountService,
```
typo

src/main/java/com/hedvig/contract/reporting/projections/MasterContractProjection.kt
```diff
@@ -202,12 +203,13 @@ class MasterContractProjection(
...
     @MetaDataValue(MetadataKeys.ORIGINAL_TIMESTAMP) originalTimeStamp: Instant?
   ) {
-    val previousOriginalTimestamp = getOriginalTimestampOfPreviousEventIfEventBreaksOriginalTimestampChain(
-      eventStore = eventStore,
-      contractId = event.contractId,
-      originalTimestamp = originalTimeStamp,
-      timestamp = timestamp
-    )
+    val previousOriginalTimestamp =
+      getOriginalTimestampOfPreviousEventIfEventBreaksOriginalTimestampChain(
```
meletios, line 207: Shall we move this to `contractStateStream` and remove the `eventStore` from here?

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
```diff
@@ -0,0 +1,136 @@
...
+    event: ContractEvent?,
+    eventTimestamp: Instant?
+  ): Boolean {
+    val correctNumber = when (event) {
+      is ContractWithActiveAgreementCreated<*> -> 1
+      else -> 2
+    }
+    if (contracts.size != correctNumber) {
+      return false
+    }
```
meletios, line 47: `correctNumber` i cannot understand why 1 and 2 is the correctNumber here. Can you rename this, to something more clear?

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
```diff
@@ -0,0 +1,136 @@
...
+import java.util.UUID
+
+@Service
+class BundleContractServiceImpl(
+  private val contractStateStream: ContractStateStream
+) : BundleContractService {
+  override fun bundleContractsMaybe(
+    contracts: List<ContractState<AgreementState>>,
+    commandGateway: CommandGateway,
+    event: ContractEvent?,
```
meletios, line 27: This should be not nullable

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
```diff
@@ -0,0 +1,136 @@
...
+    eventTimestamp: Instant?
```
meletios, line 28: Same here

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
meletios: based on the use in the saga

src/main/java/com/hedvig/contract/util/EventUtils.kt
```diff
@@ -0,0 +1,75 @@
...
+    timestamp = timestamp
+  )
+
+  if (!eventMessages.last().isOfEvent<AgreementRemoved>()) return null
+
+  val lastNonRemovalEventMessage = eventMessages.last { eventMessage: DomainEventMessage<*> ->
+    !eventMessage.isOfEvent<AgreementRemoved>()
+  }
+
+  val lastEventIsTermination = lastNonRemovalEventMessage.isOfEvent<ContractTerminated>()
```
FredrikAreschoug, line 29: isn't `lastNonRemovalEventMessage` always a boolean?

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
```diff
@@ -0,0 +1,134 @@
...
+      return false
+    }
+    val thisContract = when (event != null) {
+      true -> contractStateStream.getContractStateFromEventStream(
+        contractId = event.contractId,
+        untilTimestamp = eventTimestamp!!
+      )
+      else -> contracts.first()
+    }
+    val thatContract = contracts.first { contract -> contract.id != thisContract.id }
```
meletios, line 55: can this be null? Should we check it and return early here? 

Also i prefer firstContract secondContract rather this and that (feel free to discard this)

src/main/java/com/hedvig/contract/util/EventUtils.kt
FredrikAreschoug: am I miss understanding something? :smile:

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    return true
+  }
+
+  private fun bundleContracts(
+    contracts: List<ContractState<AgreementState>>,
+    commandGateway: CommandGateway,
+    event: ContractEvent?,
+    eventTimestamp: Instant?
```
meletios, line 89: Not-nullable or?

src/main/java/com/hedvig/contract/util/EventUtils.kt
```diff
@@ -0,0 +1,75 @@
...
+  if (originalTimestamp != null) return null
+  val eventMessages =
+    getAllEventMessages(eventStore = eventStore, aggregateIdentifier = contractId.toString()).filter { eventMessage ->
+      !eventMessage.isOfEvent<AgreementRemoved>()
+    }
+  val indexOfEvent = eventMessages.indexOfFirst { event -> event.timestamp == timestamp }
+  if (eventMessages.size > indexOfEvent + 1) {
+    val nextEventMessage = eventMessages[indexOfEvent + 1]
+    val nextEventOriginalTimestamp = nextEventMessage.metaData[MetadataKeys.ORIGINAL_TIMESTAMP] as Instant?
+    if (nextEventOriginalTimestamp != null && nextEventOriginalTimestamp.isBefore(timestamp)) {
```
FredrikAreschoug, line 53: Should we also check that `indexOfEvent - 1` is not less than 0?

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
```diff
@@ -0,0 +1,134 @@
...
+        startDate = bundleStartDate
+      )
+    )
+  }
+
+  fun getBundleStartDate(
+    thisContract: ContractState<AgreementState>,
+    thatContract: ContractState<AgreementState>,
+    event: ContractEvent?,
+    eventTimestamp: Instant?
```
meletios, line 124: Not-nullable or?

src/main/java/com/hedvig/productPricing/campaign/service/BundleDiscountService.kt
```diff
@@ -0,0 +1,21 @@
...
+import com.hedvig.productPricing.campaign.service.dto.BundleDiscount
+import com.hedvig.productPricing.campaign.service.dto.InsuranceType
+import java.time.YearMonth
+import java.util.UUID
+
+interface BundleDiscountService {
+  fun getBundleDiscountOrNullFromInsuranceTypes(insuranceTypes: List<InsuranceType>): BundleDiscount?
+  fun getBundlePremiumDiscountFromAgreement(agreement: BaseAgreement, bundledWithContractId: UUID): BundlePremiumDiscount
+  /**
+   * Use getFullDiscount to get the full discount this to be able to display monthly cost
```
FredrikAreschoug, line 14: :+1:

src/main/java/com/hedvig/productPricing/campaign/service/CampaignServiceImpl.kt
```diff
@@ -41,7 +41,7 @@ class CampaignServiceImpl(
   private val redeemedCampaignRepository: RedeemedCampaignRepository,
   private val underwriterService: UnderwriterService,
   private val premiumService: PremiumService,
-  private val bundleDiscountsService: BundleDiscountsService
+  private val bundleDiscountsService: BundleDiscountService
```
meletios, line 44: typo
```suggestion
  private val bundleDiscountService: BundleDiscountService
```

src/main/java/com/hedvig/productPricing/campaign/service/CampaignBillingServiceImpl.kt
```diff
@@ -26,7 +26,7 @@ class CampaignBillingServiceImpl(
   val campaignOwnerRepository: CampaignOwnerRepository,
   val campaignService: CampaignService,
   val premiumBillingMapper: PremiumBillingMapper,
-  val bundleDiscountsService: BundleDiscountsService
+  val bundleDiscountsService: BundleDiscountService
```
meletios, line 29: val bundleDiscountService: BundleDiscountService

src/main/java/com/hedvig/contract/reporting/projections/MasterContractProjection.kt
vonElfvin: I think this is fine since it's in `EventUtils` which makes sense to me at least

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
```diff
@@ -0,0 +1,136 @@
...
+  override fun bundleContractsMaybe(
+    contracts: List<ContractState<AgreementState>>,
+    commandGateway: CommandGateway,
+    event: ContractEvent?,
+    eventTimestamp: Instant?
+  ) {
+    if (!canBundle(contracts, event, eventTimestamp)) {
+      return
+    }
+    bundleContracts(contracts, commandGateway, event, eventTimestamp)
```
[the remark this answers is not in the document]
vonElfvin: I completely see your point but I think this is up to personal preference. I personally think early returns are very explicit and clear. Especially if we would want to do do multiple checks for example then two early returns is better than nested/longer if statements. If you strongly disagree I can change it but I personally prefer this way.

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
```diff
@@ -0,0 +1,136 @@
...
+    eventTimestamp: Instant?
+  ): Boolean {
+    val correctNumber = when (event) {
+      is ContractWithActiveAgreementCreated<*> -> 1
+      else -> 2
+    }
+    if (contracts.size != correctNumber) {
+      return false
+    }
+    val thisContract = when (event != null) {
```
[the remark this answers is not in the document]
vonElfvin: Removed it entirely now actually. Until we fix the backfill the event cannot be null regardless

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
```diff
@@ -0,0 +1,136 @@
...
+    val thatContract = when (event != null) {
+      true -> contracts.first { contract -> contract.id != thisContract.id }
+      else -> contracts.last()
+    }
+
+    val sameMember = thisContract.holderMemberId == thatContract.holderMemberId
+
+    if (!sameMember) {
+      return false
+    }
```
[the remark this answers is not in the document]
vonElfvin: Yes of course but the current one is a lot more readable IMO

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
```diff
@@ -0,0 +1,136 @@
...
+
+    val bothHasBeenActive = thisContract.hasBeenActive && thatContract.hasBeenActive
+    if (!bothHasBeenActive) {
+      return false
+    }
+
+    val bothAreActive = !thisContract.isTerminated && !thatContract.isTerminated
+    if (!bothAreActive) {
+      return false
+    }
```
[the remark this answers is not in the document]
vonElfvin: The issue is that the contract can be pending. I think clarifying that it cannot be pending makes it more explicit. I'll change the above one to that instead.

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
```diff
@@ -0,0 +1,136 @@
...
+
+    val validTypes = when {
+      thisContract is NorwegianHomeContentContractState && thatContract is NorwegianTravelContractState -> true
+      thisContract is NorwegianTravelContractState && thatContract is NorwegianHomeContentContractState -> true
+      else -> false
+    }
+    if (!validTypes) {
+      return false
+    }
```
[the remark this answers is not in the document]
vonElfvin: I personally think my implementation is a lot easier to read and understand. Also easier to add more valid cases.

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
```diff
@@ -0,0 +1,136 @@
...
+    event: ContractEvent?,
+    eventTimestamp: Instant?
+  ) {
+    val thisContract = when (event != null) {
+      true -> contractStateStream.getContractStateFromEventStream(
+        contractId = event.contractId,
+        untilTimestamp = eventTimestamp!!
+      )
+      else -> contracts.first()
+    }
```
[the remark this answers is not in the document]
vonElfvin: For now I removed the "first" thing entirely. Will probably do that later though

src/main/java/com/hedvig/contract/util/EventUtils.kt
vonElfvin: `lastNonRemovalEventMessage` is a `DomainEventMessage`. This checks if it is a termination

src/main/java/com/hedvig/contract/util/EventUtils.kt
vonElfvin: I don't remember exactly how this works but I'm pretty sure that cannot happen due to the first event being at least a contract created or something. I do however agree that it should be included and have added it.

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
vonElfvin: I would rather crash if it happened, since `canBundle` more or less ensures this cannot be null.

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
vonElfvin: Yes I changed it now, will need to make updates for the backfill though since it doesn't use events.

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
vonElfvin: Yes I changed it now, will need to make updates for the backfill though since it doesn't use events.

src/main/java/com/hedvig/contract/util/EventUtils.kt
FredrikAreschoug: aa read though ```val lastNonRemovalEventMessage = eventMessages.last { eventMessage: DomainEventMessage<*> ->
    !eventMessage.isOfEvent<AgreementRemoved>()
  }``` again it's `last` not `let` :facepalm:, disregard this comment!

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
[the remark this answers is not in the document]
FredrikAreschoug: Definitely a personal preference, I like early returns but in this case I would not use it but yeah that is my personal preference. Keep it

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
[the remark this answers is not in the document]
FredrikAreschoug: I can see your point in the `val` and not have a comment

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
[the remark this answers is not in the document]
FredrikAreschoug: If it's about the `val` yes I can see your point but the when statement is excessive in my book, but do as you please it's a suggestion

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
```diff
@@ -69,8 +65,8 @@ class BundleContractServiceImpl(
       return false
     }
 
-    val bothHasBeenActive = thisContract.hasBeenActive && thatContract.hasBeenActive
-    if (!bothHasBeenActive) {
+    val neitherIsPending = !thisContract.hasPendingAgreement && !thatContract.hasPendingAgreement
```
meletios, line 68: what about terminated?

meletios: Should we check that also?

vonElfvin: That one is below

src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
```diff
@@ -75,6 +75,11 @@ class BundleContractServiceImpl(
       return false
     }
 
+    val neitherIsBundled = !thisContract.isBundledWithOtherContract && !thatContract.isBundledWithOtherContract
+    if (!neitherIsBundled) {
+      return false
+    }
```
FredrikAreschoug, line 81: Got to love a double negative :smile:

PR #540, opened 2020-05-08
Catch `PredecessorTimelinePartWithoutEndException` and implemented `ApiErrors`

FredrikAreschoug:
fix for: https://sentry.io/organizations/hedvig/issues/1652079569/?environment=production&project=296779&referrer=alert_email


src/main/java/com/hedvig/productPricing/service/web/error/ApiError.kt
```diff
@@ -0,0 +1,6 @@
+package com.hedvig.productPricing.service.web.error
+
+data class ApiError(
```
johantjellden, line 3: Should we start having a class like this returned when status code != 2XX?

FredrikAreschoug: I think that would be super nice, guessing we would want to have some internal error codes to be able to recover errors like this. Guessing Bacoffice just blows up in this case and recover that and present it to iex would be a lot nicer!

FredrikAreschoug: But I've been thinking and a builder for `ResponseEntity` would nice to implement for this!

src/main/java/com/hedvig/common/web/error/GlobalExceptionHandler.kt
```diff
@@ -0,0 +1,27 @@
...
+  fun handleException(exception: PredecessorTimelinePartWithoutEndException, request: WebRequest): ResponseEntity<ApiErrorResponse> {
+    return ResponseEntity(
+      ApiErrorResponse(
+        ApiError(
+          code = predecessorTimelinePartWithoutEndCode,
+          message = predecessorTimelinePartWithoutEndMessage
+        )
+      ),
+      HttpHeaders(),
+      HttpStatus.CONFLICT
```
johantjellden, line 24: 👍

PR #546, opened 2020-05-13
Feature/bundle backfill

vonElfvin:
Will need to fix the PR comments, merge and test the `feature/bundle-discount` branch first.


src/main/java/com/hedvig/contract/service/BundleContractServiceImpl.kt
```diff
@@ -93,24 +154,34 @@ class BundleContractServiceImpl(
...
-
+    val thatContract = contracts.first { contract -> contract.id != event.contractId }
     val bundleStartDate = getBundleStartDate(thisContract, thatContract, event, eventTimestamp)
+    bundleContracts(thisContract, thatContract, commandGateway, bundleStartDate)
+  }
 
+  private fun bundleContracts(
+    thisContract: ContractState<AgreementState>,
+    thatContract: ContractState<AgreementState>,
+    commandGateway: CommandGateway,
```
meletios, line 165: Why are you passing around the commandGateway and not injecting in the constructor of the class?

src/main/java/com/hedvig/contract/web/InternalBundleController.kt
```diff
@@ -0,0 +1,19 @@
...
+
+@RestController
+@RequestMapping("/_/bundle")
+class InternalBundleController(
+  private val bundleContractService: BundleContractService,
+  private val commandGateway: CommandGateway
+) {
+  @PostMapping("/backfill/norway")
+  fun backfillNorwegianBundles() {
+    bundleContractService.backfillMissingNorwegianBundles(commandGateway)
```
meletios, line 17: I guess the same comment here

PR #555, opened 2020-05-19
Feature/get all partner campaign codes

AlexSJ1990:
returns campaign code data for all partner campaigns - this will be used to show info re campaigns in the tools section of hope. This will not include referrals.


src/main/java/com/hedvig/productPricing/campaign/service/CampaignSearchService.kt
```diff
@@ -1,17 +1,21 @@
...
 @Service
 class CampaignSearchService(
-  val campaignRepository: CampaignRepository
+  val campaignRepository: CampaignRepository,
+  val campaignOwnerRespository: CampaignOwnerRepository
 ) {
-  fun searchForPartnerCampaigns(code: String? = null): List<PartnerCampaignSearchResponse> {
-    val campaignSearchResults = campaignRepository.findAllPartnerCampaigns(code = code)
+  fun searchForPartnerCampaigns(code: String? = null, dateToday: LocalDate? = null): List<PartnerCampaignSearchResponse> {
+    val campaignSearchResults = campaignRepository.findAllPartnerCampaigns("")
```
johantjellden: We need to fix this.

src/main/java/com/hedvig/productPricing/campaign/domain/CampaignRepository.kt
```diff
@@ -19,8 +18,8 @@ interface CampaignRepository : CrudRepository<Campaign, UUID> {
     join Partner p on p.ownerId = c.campaignOwner
     where lower(c.code) like ('%' || lower(:code) || '%')
     and lower(p.partnerId) like ('%' || lower(:partnerId) || '%')
-    and (:activeFrom is null or c.validFrom is null or :activeFrom >= c.validFrom)
-    and (:activeTo is null or c.validUntil is null or :activeTo <= c.validUntil)
+    and (CAST(:activeFrom as date) is null or c.validFrom is null or :activeFrom >= c.validFrom)
```
FredrikAreschoug, line 21: shouldn't we cast it when checking if it before or after validFrom/validUntil, instead of casting when checking if its null or not?

PR #559, opened 2020-05-26
Increase base factor instead of adding inflation correction

vonElfvin:


src/main/java/com/hedvig/productPricing/pricing/PricingEngineHedvig_v1.java
```diff
@@ -36,7 +36,7 @@
   static enum TravelDestination {EUROPE, NORTH_AMERICA, OTHER};
   static enum ObjectType {MUSIC_INSTRUMENT, WATCH, JEWELRY, CAMERA, ART, OTHER};
 
-  private static double price_factor_apartment = 0.79d;
+  private static double price_factor_apartment = 0.83d;
```
vonElfvin, line 39: 0.79*1.05=0.8295

PR #560, opened 2020-05-26
Add charge info to swedish and norwegian insurance certificates

vonElfvin:
Add sentence about when a charge is performed on the insurance certificate. The requirement comes from HDI and copy from Sara/Ermir


src/test/java/com/hedvig/productPricing/service/service/certificates/InsuranceCertificateTemplateTest.kt
```diff
@@ -1,6 +1,7 @@
 package com.hedvig.productPricing.service.service.certificates
 
 import com.hedvig.contract.shared.certificates.MemberInfo
+import com.hedvig.contract.shared.certificates.mock.MockNorwegianInsuranceCertificate
```
palmenhq, line 4: ?

src/main/java/com/hedvig/contract/shared/certificates/mock/MockInsuranceCertificateController.kt
```diff
@@ -0,0 +1,37 @@
+package com.hedvig.contract.shared.certificates.mock
+
+import com.hedvig.contract.service.InsuranceCertificateGenerator
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+@RequestMapping("/_/mock/insurance/certificate")
```
palmenhq, line 9: Maybe we should kebab case this, and try to shoe horn it to some kind of resource rather than using `/` and verbs. Something like `/_/insurance-certificates/mocks` :)

src/test/java/com/hedvig/productPricing/service/service/certificates/InsuranceCertificateTemplateTest.kt
vonElfvin: I moved the file to be usable outside test, then I had to import it here

PR #563, opened 2020-06-03
Redeem v2 iteration

FredrikAreschoug:
Mostly mapping `CampaignNotFoundExceptions` and `RedeemDeclinedException` to graphql union.

But while building the test I also found that we didn't use valid until on campaigns so added validation for that on redeeming:smile:


src/test/java/com/hedvig/productPricing/campaign/service/CampaignServiceImplTest.kt
```diff
@@ -41,8 +45,9 @@ class CampaignServiceImplTest {
     underwriterService = mockk()
     premiumService = mockk()
     bundleDiscountService = mockk()
+    contractService = mockk()
 
-    classUnderTest = CampaignServiceImpl(campaignOwnerRepository, insuranceService, redeemedCampaignRepository, underwriterService, premiumService, bundleDiscountService)
+    classUnderTest = CampaignServiceImpl(campaignOwnerRepository, insuranceService, redeemedCampaignRepository, underwriterService, premiumService, bundleDiscountService, contractService)
```
AlexSJ1990, line 50: remove contractService?

PR #565, opened 2020-06-04
change column type from char var 255 to text

akashpatl:


src/main/java/com/hedvig/contract/shared/models/ContractState.kt
```diff
@@ -41,18 +41,7 @@ import java.time.LocalDate
...
-import javax.persistence.Enumerated
-import javax.persistence.FetchType
-import javax.persistence.Id
-import javax.persistence.Inheritance
-import javax.persistence.InheritanceType
-import javax.persistence.JoinColumn
-import javax.persistence.JoinTable
-import javax.persistence.OneToMany
-import javax.persistence.OneToOne
+import javax.persistence.*
```
meletios, line 44: Me no like dis 😅

vonElfvin: But I doooo

PR #567, opened 2020-06-05
Create members in oracle when a contract is created

vonElfvin:
- Creates a member in oracle when a contract is created
- Dependent on oracle being out in both staging and prod
- Oracle does not throw exceptions for when you try to create a member twice (i.e. does not crash when a bundle is created for example)


src/main/java/com/hedvig/productPricing/service/serviceIntegration/oracle/OracleEventListener.kt
```diff
@@ -0,0 +1,44 @@
...
+    } catch (exception: Exception) {
+      logger.error("Unable to create member in Oracle (memberId=${event.holderMemberId}", exception)
+      throw exception
+    }
+  }
+
+  @EventHandler
+  fun on(event: ContractWithPendingAgreementCreated<*>) {
+    try {
+      logger.info("Created member in Oracle (memberId=${event.holderMemberId})")
```
FredrikAreschoug, line 32: Should move this one line down right?

vonElfvin: That is true! Nice catch

PR #574, opened 2020-06-12
Feature/notify when a renewal is queued

meletios:


src/main/java/com/hedvig/contract/customerio/Sweden.kt
```diff
@@ -65,13 +66,13 @@ class Sweden(
...
-      eventId = this.eventId,
-      memberId = this.contract.holderMemberId
-    )
+  override fun onContractRenewalQueued(contract: ContractState<AgreementState>?) {
+    if (contract != null) {
+      val data = mapOf("renewal_date" to contract.renewal!!.renewalDate.toString())
+      notificationService.updateCustomer(contract.holderMemberId, data)
+    } else {
+      logger.error("Could not find a contract for event ContractRenewalQueued")
+    }
```
FredrikAreschoug, line 75: I prefer making the null check early
```suggestion
  override fun onContractRenewalQueued(contract: ContractState<AgreementState>) {
      val data = mapOf("renewal_date" to contract.renewal!!.renewalDate.toString())
      notificationService.updateCustomer(contract.holderMemberId, data)
```

src/main/java/com/hedvig/contract/customerio/EventListener.kt
```diff
@@ -244,10 +244,11 @@ class EventListener(
     eventMessage: DomainEventMessage<Any>
   ) {
     onlyRunIfEventIsNewerThanComponent(timestamp, event) {
+      val contract = contractStateRepository.findByIdOrNull(event.contractId)
       applyStrategyToContract(
         event.contractId,
         eventId = eventMessage.identifier
-      ) { it.onContractRenewalQueued(event.draftCertificate.inceptionDate!!) }
+      ) { it.onContractRenewalQueued(contract) }
```
FredrikAreschoug, line 251: I prefer making the null check early
```suggestion
      ) { 
      contract?.let { contract -> 
        it.onContractRenewalQueued(contract)
         } ?: logger.error("Could not find a contract for event ContractRenewalQueued")
      }
```

PR #576, opened 2020-06-12
Solve issues with swticher emails query

vonElfvin:
In this PR:
- Rename/extract some files for clarity
- Solve 3 issues in the query for getting switcher emails (both to send and have sent):
1. Sort emails by DESC so that if a member has for some reason received multiple emails for one contract type we return the latest and not the oldest (the bug Fredrik and Alex noticed when testing this)
2. In addition to excluding contracts with active agreements we also exclude contracts that are terminated
3. When performing the LEFT join on contracts we also check if the lineOfBusiness for the email matches the type of contract. With combo we otherwise risk joining on the wrong contract.


src/main/java/com/hedvig/contract/shared/switchers/SwitchableSwitcherEmailEventListener.kt
```diff
@@ -1,44 +1,40 @@
...
-      is NorwegianTravelLineOfBusiness -> SwitcherLineOfBusiness.NORWEGIAN_TRAVEL
-      is SwedishApartmentLineOfBusiness -> SwitcherLineOfBusiness.SWEDISH_APARTMENT
-      is SwedishHouseLineOfBusiness -> SwitcherLineOfBusiness.SWEDISH_HOUSE
-      else -> throw IllegalStateException("Cannot handle line of business $lineOfBusiness for memberId: ${event.holderMemberId}")
+    val lineOfBusiness = when (event.agreement) {
+      is NorwegianHomeContentAgreement -> SwitcherLineOfBusiness.NORWEGIAN_HOME_CONTENT
+      is NorwegianTravelAgreement -> SwitcherLineOfBusiness.NORWEGIAN_TRAVEL
+      is SwedishApartmentAgreement -> SwitcherLineOfBusiness.SWEDISH_APARTMENT
+      is SwedishHouseAgreement -> SwitcherLineOfBusiness.SWEDISH_HOUSE
+      else -> throw IllegalStateException("Cannot handle switcher line of business for agreement of type ${event.agreement::class.simpleName} (memberId=${event.holderMemberId})")
```
vonElfvin, line 28: Also, since house's lineOfBusiness is always null I think this is a more robust solution. I'm unsure how `is SwedishHouseLineOfBusiness` will work since it is null

src/main/java/com/hedvig/contract/shared/aggregates/BaseAgreement.kt
```diff
@@ -8,6 +8,7 @@ import com.hedvig.contract.shared.models.ContractState
 import com.hedvig.contract.shared.models.objects.InsuranceCertificate
 import com.hedvig.contract.shared.stereotypes.AgreementStereotype
 import com.hedvig.contract.shared.stereotypes.LineOfBusinessStereotype
+import com.hedvig.contract.shared.switchers.SwitcherLineOfBusiness
```
palmenhq, line 11: is this intentional?

vonElfvin: Not it is not, I'll keep it now since pp takes 10 minutes to build. Have removed it locally though so eventually it will be gone 🤦

PR #585, opened 2020-06-29
fix billing bug

meletios:
add a fix for  currency missmatc on billing causing norwegian discounts to be null


src/main/java/com/hedvig/productPricing/campaign/domain/model/Member.kt
```diff
@@ -303,7 +303,7 @@ open class Member(
       DateUtil.toLocalDate(redeemedCampaign.activatedAt!!),
       redeemedCampaign.endDateOfDiscount()
     )
-    if (remainingToZero == null || remainingToZero <= Money.of(0, "SEK")) {
+    if (remainingToZero == null || remainingToZero.isNegativeOrZero) {
```
vonElfvin, line 306: Please ensure this is present for the non free month discount as well, I presume they use the same function with a discount of 100% but just to double check :)

PR #587, opened 2020-06-29
correct missed campaigns

meletios:
Correct missed campaigns by setting up the activation date on first of July. This applies only to switchers until end of June 2020


src/main/java/com/hedvig/productPricing/service/configuration/AxonConfiguration.java
```diff
@@ -94,6 +94,15 @@ public void campaignConfiguration(EventProcessingConfiguration configuration) {
     );
   }
 
+  @Autowired
+  public void campaignCorrectionConfiguration(EventProcessingConfiguration configuration) {
+    configuration.registerTrackingEventProcessor(
+      "CampaignCorrectionEventListener",
+      x ->
+        TrackingEventProcessorConfiguration.forSingleThreadedProcessing()
+          .andInitialTrackingToken(StreamableMessageSource::createHeadToken)
```
vonElfvin, line 103:
```suggestion
          .andInitialTrackingToken(StreamableMessageSource::createTailToken)
```

src/main/java/com/hedvig/productPricing/service/query/CampaignCorrectionEventListener.kt
```diff
@@ -0,0 +1,52 @@
...
+import com.hedvig.productPricing.campaign.util.DateUtil
+import org.axonframework.config.ProcessingGroup
+import org.axonframework.eventhandling.EventHandler
+import org.axonframework.eventhandling.Timestamp
+import org.springframework.stereotype.Component
+import java.time.Instant
+import java.time.LocalDate
+
+@Component
+@ProcessingGroup("CampaignCorrectionEventListener")
```
vonElfvin, line 19: Maybee add a profile as well

src/main/java/com/hedvig/productPricing/service/query/CampaignCorrectionEventListener.kt
```diff
@@ -0,0 +1,52 @@
...
+  val campaignOwnerRepository: CampaignOwnerRepository
+) {
+
+  @EventHandler
+  fun on(e: PendingAgreementActivated<BaseAgreement>, @Timestamp timestamp: Instant) {
+    if (!timestamp.isBefore(JUNE_30_2020)) {
+      return
+    }
+
+    val memberId = contractStateRepository.findById(e.contractId).get().holderMemberId
```
vonElfvin, line 32: We should probable get the whole contract here and do a bit more actually. Either:
1. If the contract is active today, activate as you do now
2. If the contract is terminated, do nothing
3. If the contract is active in the future, activate from that date

src/main/java/com/hedvig/productPricing/service/query/CampaignCorrectionEventListener.kt
meletios: Ah, its too much. I will discard this.

src/main/java/com/hedvig/productPricing/service/configuration/AxonConfiguration.java
meletios: God dammit, i am always confused about it!

PR #590, opened 2020-07-02
using campaign service get total instead of premium service

FredrikAreschoug:
One last bug with the premium in `calculateCostReducedIndefinite`


FredrikAreschoug:
But I did, or did I miss it in the test?

PR #592, opened 2020-07-06
Feature/endpoint batch change member campaign code

FredrikAreschoug:
[Notion ticket](https://www.notion.so/hedviginsurance/a1d3df55acd843cda523589ff05267b2?v=2602c26d8d5540c68568c668cf4b9c95&p=51c5f9585b174dad81049afcd5e03281). Added some validation before changing the campaign code mostly because I don't want to create new bugs and did not have time to look into the consequences if the checks are not there.


src/main/java/com/hedvig/productPricing/campaign/service/UpdateCampaignsServiceImpl.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    if (!lowerCaseCodes.hasOnlyUniqueItems()) {
+      return BatchChangeCampaignsResult.CodeSetIsNotUnique
+    }
+
+    lowerCaseCodes.forEach { code ->
+      if (campaignCodeRepository.existsById(code)) {
+        return BatchChangeCampaignsResult.OneOrMoreCampaignCodesAreTaken(code)
+      }
+    }
```
meletios, line 27: Not sure about the logic that we want to have here BUT, isnt it better to check all the codes and return a list of taken codes, instead of early return on the first one?

meletios: Shall we change the non-taken codes also? (Probably it will be wrong but just a suggestion)

FredrikAreschoug: Yeah I could return all was just lazy, this is an endpoint for until we have something better in hope so I rather check all beforehand and if fails I don't want any changes to take place

PR #610, opened 2020-07-22
decoupled campaign service form contracts

FredrikAreschoug:
I made a stab at decoupling campaigns service from contracts. I think we should also decouple graphql to have a product-pricing graphl instead of two for one service wich creates some spagetti code, also we should move out billings from campaigns. But let's do that later on


src/main/java/com/hedvig/productPricing/campaign/integration/contract/ContractMapper.kt
```diff
@@ -0,0 +1,12 @@
+package com.hedvig.productPricing.campaign.integration.contract
+
+import com.hedvig.productPricing.campaign.integration.contract.model.ContractForBundling
+import com.hedvig.productPricing.campaign.integration.contract.model.Market
+import com.hedvig.productPricing.campaign.integration.contract.model.MarketCurrency
+
+interface ContractMapper {
+  fun getMarketFromMember(memberId: String): Market?
+  fun getMarketCurrency(memberId: String): MarketCurrency?
+  fun getContractForBundlings(memberId: String): List<ContractForBundling>
```
AlexSJ1990, line 10: getContractForBundle maybe?

src/main/java/com/hedvig/productPricing/campaign/integration/contract/model/ContractForBundling.kt
```diff
@@ -0,0 +1,10 @@
+package com.hedvig.productPricing.campaign.integration.contract.model
+
+import com.hedvig.productPricing.campaign.service.dto.InsuranceType
+import java.time.LocalDate
+
+data class ContractForBundling(
```
AlexSJ1990, line 6: ContractForBundle

src/main/java/com/hedvig/productPricing/campaign/service/ReferralServiceImpl.kt
```diff
@@ -106,7 +105,7 @@ constructor(
   }
 
   private fun calculateCostReducedIndefiniteDiscount(member: Member): InsuranceCost? =
-    premiumService.getTotalPremiumOfContractsWithActiveActiveInFutureOrPendingAgreementsByHolderMemberId(member.memberId)?.let { premium ->
+    premiumBillingMapper.getCurrentOrUpCommingTotalPremium(member.memberId)?.let { premium ->
```
AlexSJ1990, line 108: getCurrentOrUpcomingTotalPremium

src/test/java/com/hedvig/contract/testUtils/builder/a.kt
```diff
@@ -71,6 +75,29 @@ class a {
       UUID.randomUUID(),
       LocalDate.now()
     )
+
+    fun swedishApartmentContractWhitLineOfBusiness(lineOfBusiness: SwedishApartmentLineOfBusiness) =
```
AlexSJ1990, line 79: `swedishApartmentContractWithLineOfBusiness`

src/main/java/com/hedvig/productPricing/campaign/integration/contract/ContractMapper.kt
```diff
@@ -0,0 +1,12 @@
+package com.hedvig.productPricing.campaign.integration.contract
+
+import com.hedvig.productPricing.campaign.integration.contract.model.ContractForBundling
+import com.hedvig.productPricing.campaign.integration.contract.model.Market
+import com.hedvig.productPricing.campaign.integration.contract.model.MarketCurrency
+
+interface ContractMapper {
+  fun getMarketFromMember(memberId: String): Market?
```
AlexSJ1990, line 8: getMarketForMember ?

src/main/java/com/hedvig/productPricing/campaign/integration/contract/ContractMapper.kt
FredrikAreschoug: Maybe it should be `getContractsForBundling` :thinking:  but I like bundling since its more a data class for bundling which bundle service is taking care of

src/main/java/com/hedvig/common/Extensions.kt
```diff
@@ -1,4 +1,4 @@
-package com.hedvig.contract.customerio
+package com.hedvig.common
```
johantjellden, line 1: Nice 👍 Do I sense a proper library?

FredrikAreschoug: We are leading up to that yes :slightly_smiling_face:

PR #611, opened 2020-07-31
Include voucher cost deduction in billing calculation

FredrikAreschoug:
While browsing around in this repo I released that partner cost deduction is not included in billing. This is the quick solution to fix that but I think we should at some point extract and rewrite billing to make sure that no redeemed campaign will ever be left out of billing. And don't forget to merge and deploy [this one](https://github.com/HedvigInsurance/account-service/pull/90)


src/main/java/com/hedvig/productPricing/campaign/domain/model/Member.kt
```diff
@@ -307,5 +309,29 @@ open class Member(
 
     return Pair(remainingToZero, redeemedCampaign.campaign.code.toUpperCase())
   }
+
+  fun getCostDeductionVoucherCampaigns(month: YearMonth): List<Pair<MonetaryAmount, String>> {
```
AlexSJ1990, line 313: getCostDeductionForVoucherCampaigns maybe?

FredrikAreschoug: then it should be more of `getCostDeductionVoucherCampaignsForMonth` But I don't know if I think that clarifies anything since it the only param

PR #616, opened 2020-08-26
Apply bundle discount on price gross, not as discount

vonElfvin:


vonElfvin:
This will most likely not solve the real issue. Closing until we find the actual root cause.

PR #619, opened 2020-09-09
Add Profile to SwitchableSwitcherEmailEventListener

vonElfvin:
We need this out ASAP to stop queueing switcher emails


johantjellden (approved):

vonElfvin:
After looking into the code I noticed we actually don't need it 🤷 I turned off the cronjob instead

PR #620, opened 2020-09-09
Mask norwegian ssn for switching emails

vonElfvin:


src/main/java/com/hedvig/contract/shared/switchers/mail/SwitcherCancellationMailerImpl.kt
```diff
@@ -47,7 +49,7 @@ class SwitcherCancellationMailerImpl(
       "for member ${norwegianEmailData.memberId}...")
 
     val context = VelocityContext()
-    context.put("ssn", norwegianEmailData.ssn)
+    context.put("ssn", maskSsn(norwegianEmailData.ssn, Market.NORWAY))
```
johantjellden, line 52: Should we not mask the swedish ones as well, on line 35.

vonElfvin: Nope only Norway for now

johantjellden (approved):

PR #621, opened 2020-09-10
open up referrals for norway

meletios:


src/main/java/com/hedvig/productPricing/campaign/service/ReferralServiceImpl.kt
```diff
@@ -174,26 +178,9 @@ constructor(
 
   override fun eligibleForReferral(memberId: String): Boolean {
     return when (contractMapper.getMarketForMember(memberId)) {
-      Market.SWEDEN -> {
+      Market.SWEDEN, Market.NORWAY  -> {
         true
       }
```
johantjellden, line 183: So we are removing the limited referral thing? :)
Is there any more code we could possibly remove as well?

akashpatl (dismissed):
If this is the only thing required for everyone in Norway to have access to Forever then it looks good. So just to be clear Norway should be exactly as Sweden.

src/main/java/com/hedvig/productPricing/campaign/service/ReferralServiceImpl.kt
```diff
@@ -174,26 +178,9 @@ constructor(
 
   override fun eligibleForReferral(memberId: String): Boolean {
```
vonElfvin, line 179: 🅱️ ig 🧠 
```suggestion
  override fun eligibleForReferral(memberId: String): Boolean = true
```

meletios: Screenshot 2020-09-11 at 14 33 42
😂

vonElfvin: Or, why not just remove the "eligibleForReferral" check entirely and just not check it?

meletios: we need to keep it, since its used by the APP

src/main/java/com/hedvig/productPricing/campaign/service/ReferralServiceImpl.kt
meletios: Nope, only this one

meletios: as far as i know

PR #624, opened 2020-09-18
Fix Norwegian Switcher Emails

vonElfvin:
Include memberId and cancellationDate for switcher emails in Norway, do not require agreementId for activation of Pending Agreements


AlexSJ1990 (approved):
Looks nice!

PR #626, opened 2020-09-27
refactor to use product-pricing-objects

AlexSJ1990:
This pr adds splits product-pricing into two different modules.

Product-pricing-objects 
Product-pricing-service

Product-pricing-objects is a library which will be used externally by other services (e.g. underwriter) so far it consists of dto's and enums related to agreements and contracts - over time we can add further relevant dto's and enums.

Any dto or enum from the external product-pricing-object library will only be allowed to get as far as the controller layer within product-pricing-service. We then have explicitly mapped to internal dto's / enums. The mappers are fairly large in this pr and going forward we can split into smaller mappers but at the moment the thought is to have a separate Ingoing and Outgoing mapper for each type of mapper for clarity.


product-pricing-service/src/main/java/com/hedvig/productPricing/service/web/dto/InsuranceDTO.java
```diff
@@ -108,11 +118,11 @@ public InsuranceDTO(Clock clock, ProductEntity product, List<PerilEntity> perils
     this.extraBuildings = product.extraBuildings != null
       ? product.extraBuildings.stream()
       .map(extraBuilding ->
-        ExtraBuildingDto.Companion.from(
-          extraBuilding,
-          localizationService,
-          locale
-        )
+          contractToProductRepository.getExtraBuildingDto(
```
johantjellden, line 121: For me repositories are used just to talk to the database and the fact that we have to inject hem here just to be map the type is a bit weird.

I would move the `getExtraBuildingDto` to a dedicated mapper class or perhaps leave it here in the controller for now.

product-pricing-service/src/main/java/com/hedvig/productPricing/service/web/dto/InsuranceStatusDTO.java
```diff
@@ -53,19 +55,24 @@
   @JsonProperty("isSubleted")
   Boolean isSubleted;
 
-  public InsuranceStatusDTO(ProductEntity productEntity) {
-    this(productEntity, new ArrayList<>());
+  public InsuranceStatusDTO(ProductEntity productEntity, ContractToProductRepository contractToProductRepository) {
```
johantjellden, line 58: Related to the other comment, we should not really need to inject repositories into DTOs :)

product-pricing-service/src/main/java/com/hedvig/contract/graphql/types/AgreementStatusType.kt
```diff
@@ -8,6 +8,8 @@ enum class AgreementStatusType {
   ACTIVE,
   TERMINATED;
 
+//  what about graphql types ??? these are external ??
```
johantjellden, line 11: These types are external but are already exposed though our graphql schema.

product-pricing-service/src/main/java/com/hedvig/contract/shared/models/ContractState.kt
```diff
@@ -83,7 +84,7 @@ abstract class ContractState<TAgreement : AgreementState>(
 
   @Suppress("UNCHECKED_CAST")
   val status: ContractStatus
-    get() = ContractStatus.createStatus(this as ContractState<AgreementState>)
+    get() = getStatus()
```
johantjellden, line 87: Drop the `@Suppress("UNCHECKED_CAST")` it is not needed anymore.

Will hibernate create a column from any of these functions?

johantjellden: Seems like @Id on primary constructor property will default to @field:Id so this should be safe. 
Which can be verified by looking in the DB :)

product-pricing-service/src/main/java/com/hedvig/productPricing/campaign/web/MemberCampaignController.kt
```diff
@@ -100,14 +102,15 @@ class MemberCampaignController(
     }
   }
 
+//  todo: show this to TJ - do the mapping in the controller layer as much as possible?
```
FredrikAreschoug, line 105: Should TJ look at it? or should we remove the comment?

product-pricing-service/src/main/java/com/hedvig/contract/web/dtos/manual/ManualReplaceAgreementRequest.kt
```diff
@@ -45,6 +46,8 @@ data class ReplaceHouseManualAgreementRequest(
   val yearOfConstruction: Int,
   val numberOfBathrooms: Int = 0,
   val extraBuildings: MutableSet<ExtraBuilding>?,
+// todo: test me
```
FredrikAreschoug, line 49: Tested?

product-pricing-service/src/main/java/com/hedvig/contract/web/InternalContractController.kt
```diff
@@ -107,18 +119,20 @@ class InternalContractController(
...
     val contractMarketInfo = contractService.getContractMarketInfoByHolderMemberId(memberId)
     return if (contractMarketInfo != null) {
-      ResponseEntity.ok(contractMarketInfo)
+      ResponseEntity.ok(outgoingDtoMapper.toContractMarketInfoDto(contractMarketInfo))
     } else {
       ResponseEntity.notFound().build()
     }
   }
 
+//TODO: test this endpoint
```
FredrikAreschoug, line 135: Tested?

product-pricing-service/src/main/java/com/hedvig/contract/web/InternalContractController.kt
```diff
@@ -107,18 +119,20 @@ class InternalContractController(
     return ResponseEntity.ok(insurableLimits)
   }
 
+//  todo: show TJ should map in the controller layer
```
FredrikAreschoug, line 122: Is this done with the changes on line 129

product-pricing-service/src/main/java/com/hedvig/contract/web/InternalAgreementController.kt
```diff
@@ -63,8 +79,11 @@ class InternalAgreementController(
...
   }
 
+//  todo: test this endpoint
   @GetMapping("/{agreementId}")
   fun getAgreement(@PathVariable agreementId: UUID): ResponseEntity<Agreement> {
-    return ResponseEntity.ok(agreementMapper.toAgreement(agreementService.getAgreementFromAgreementId(agreementId)))
+  return ResponseEntity.ok(outgoingDtoMapper.toAgreementDto(agreementService.getAgreementFromAgreementId(agreementId)))
   }
+
+//  write a test for agreement mapper
```
FredrikAreschoug, line 88: Left over?

product-pricing-service/src/main/java/com/hedvig/contract/web/InternalAgreementController.kt
```diff
@@ -63,8 +79,11 @@ class InternalAgreementController(
     return ResponseEntity.noContent().build()
   }
 
+//  todo: test this endpoint
```
FredrikAreschoug, line 82: do or remove?

product-pricing-service/src/main/java/com/hedvig/contract/shared/models/repositories/ContractToProductRepository.kt
```diff
@@ -192,4 +194,10 @@ class ContractToProductRepository(
       )
     }
   }
+
+  fun getExtraBuildingDto(oldExtraBuilding: ExtraBuilding): ExtraBuildingDto =
+    com.hedvig.contract.shared.models.objects.ExtraBuilding.from(oldExtraBuilding).toExtraBuildingDtoWithDefaultDisplayName()
```
FredrikAreschoug, line 199: that's a long line maybe import like this: 
`import com.hedvig.contract.shared.models.objects.ExtraBuilding as ExtraBuildingShared`

product-pricing-service/src/main/java/com/hedvig/contract/service/dtos/ContractMarketInfo.kt
```diff
@@ -1,8 +1,9 @@
-package com.hedvig.contract.web.dtos
+package com.hedvig.contract.service.dtos
 
 import com.hedvig.contract.shared.models.objects.Market
 import javax.money.CurrencyUnit
 
+//move all that will be mapped in controller
```
FredrikAreschoug, line 6: Todo?

product-pricing-service/src/main/java/com/hedvig/contract/mappers/IncomingDtoMapper.kt
```diff
@@ -0,0 +1,140 @@
...
+import com.hedvig.contract.type.norwegianTravel.aggregates.NorwegianTravelAgreement
+import com.hedvig.contract.type.swedishApartment.aggregates.SwedishApartmentAgreement
+import com.hedvig.contract.type.swedishHouse.aggregates.SwedishHouseAgreement
+import org.javamoney.moneta.Money
+import org.springframework.stereotype.Component
+import java.util.UUID
+import com.hedvig.productPricingObjects.dtos.AgreementQuote as AgreementQuoteDto
+
+@Component
+class IncomingDtoMapper {
```
FredrikAreschoug, line 18: Is the plan for this mapper to map anything else that Quotes and Agreements? I feel like the naming is really broad while the implementation is quite narrow. Should we call it `AgreementQuoteMapper`

product-pricing-service/src/main/java/com/hedvig/contract/mappers/OutgoingDtoMapper.kt
```diff
@@ -1,71 +1,78 @@
...
-import com.hedvig.graphql.commons.type.MonetaryAmountV2
+import com.hedvig.productPricingObjects.dtos.Contract
+import com.hedvig.productPricingObjects.dtos.Renewal
+import com.hedvig.productPricingObjects.enums.TypeOfContract
 import org.springframework.stereotype.Component
-import java.util.Locale
+import com.hedvig.productPricingObjects.dtos.Agreement as AgreementDto
+import com.hedvig.productPricingObjects.dtos.GenericAgreement as GenericAgreementDto
+
+//  TODO: write test
```
FredrikAreschoug, line 22: Something we should do?

product-pricing-service/src/main/java/com/hedvig/contract/mappers/OutgoingDtoMapper.kt
```diff
@@ -1,71 +1,78 @@
...
+  )
+}
+
+fun toNorwegianHomeContentLineOfBusiness(norwegianHomeContentLineOfBusinessObject: com.hedvig.productPricingObjects.enums.NorwegianHomeContentLineOfBusiness): NorwegianHomeContentLineOfBusiness {
+  return NorwegianHomeContentLineOfBusiness.valueOf(norwegianHomeContentLineOfBusinessObject.name)
+}
 
 @Component
-class AgreementMapper(
+class OutgoingDtoMapper(
```
FredrikAreschoug, line 50: I don't know If I like this naming `OutgoingDtoMapper` is anything `OutgoingAgreementMapper` and separate the contract mapping to `OutgoingContractMapper`. Feels like we are building a humongous mapper class otherwise

FredrikAreschoug (commented):
I always find it hard to review prs in this size please let me know if there are som classes I really should look into

product-pricing-service/src/main/java/com/hedvig/contract/mappers/OutgoingDtoMapper.kt
johantjellden: Yes you might be right but before we had small mappers and it was really hard to know which types it mapped. In a way i don't know if we really care. We have agreements on both sides so I felt that naming them with their purpose was better than naming them with their mapping type.

Also when we do the outward mapping we need the url signer and the localize client.
So I thought that we have our inbound and outbound mappers and when they get to big we can split them up internally.

FredrikAreschoug: Okay but I still prefer separation and there is nothing that says that we cant have purpose and type in the name `OutgoingAgreementMapper`. If it is hard to understand what the mappers map, maybe we should change the project structure and not having one folder with services and one with mappers. thay ould be organized by feature. You can decide but having one outgoing and one incoming mapper for everything in contracts is way to big in my opinion.

product-pricing-service/src/main/java/com/hedvig/contract/mappers/IncomingDtoMapper.kt
vonElfvin: I agree

product-pricing-service/src/main/java/com/hedvig/contract/mappers/OutgoingDtoMapper.kt
vonElfvin: I agree with Fredrik here. Incoming and outgoing makes sense but I personally dislike having one for **ALL** dtos. `OutgoingAgreementMapper` is a good path to go IMO.

FredrikAreschoug (dismissed):
Looks good to me, not going to review all of this again. We went through the code yesterday :slightly_smiling_face:

FredrikAreschoug (approved):

PR #627, opened 2020-09-28
Fix/get bundel discount for start month

FredrikAreschoug:
I heard that this project is getting some improvements so didn't want to make a big mess while fixing this rather small bug. But when @AlexSJ1990 and @vonElfvin is done with there clean up I vote for extracting graphQL resolvers to its own module. It's a long-overdue improvement in my opinion :slightly_smiling_face:


src/main/java/com/hedvig/contract/util/ContractUtils.kt
```diff
@@ -2,6 +2,19 @@ package com.hedvig.contract.util
...
+  this.map { contractState -> contractState.pendingOrCurrentOrNextOrLastAgreement }
+    .filter { agreementState ->
+      when (AgreementStatus.createStatus(agreementState)) {
+        AgreementStatus.ACTIVE_IN_PAST,
+        AgreementStatus.TERMINATED -> false
+        AgreementStatus.PENDING,
+        AgreementStatus.ACTIVE_IN_FUTURE,
+        AgreementStatus.ACTIVE -> true
+      }
+    }
```
vonElfvin, line 20: Should probably look at `ContractStatus` instead of AgreementStatus. There it would be as easy as: Terminated = Terminated, everything else = Active

src/main/java/com/hedvig/contract/graphql/resolvers/ContractQuery.kt
```diff
@@ -40,7 +41,18 @@ class ContractQuery(
...
+    //fixme: This is a lot of code in the wrong place but lets fix that by breaking out graphQL from contracts and campaign to it own module
+    val firstStartDate = contractService.getContractStatesByHolderMemberId(memberId)
+      .filterActivePendingOrActiveInFutureAgreements()
+      .filter { it.from != null }
+      .minBy { it.from!! }?.from
+
+    val periodToCheck = firstStartDate?.let {
+      val startMonth = YearMonth.from(it)
+      if (startMonth.isBefore(YearMonth.now())) YearMonth.now() else startMonth
+    } ?: YearMonth.now()
```
vonElfvin, line 53: What would happen if we just did "YearMonth.now()" straight up?

vonElfvin (commented):
Great catch! Some thoughts :)

src/main/java/com/hedvig/contract/graphql/resolvers/ContractQuery.kt
vonElfvin: I think we should use masterInception which is a set concept. Also a minor refactoring of the date calculation using Kotlin's `max`

```suggestion
    val firstMasterInceptionMaybe = contractService.getContractStatesByHolderMemberId(memberId)
      .filterActiveOrPendingContracts() // based on ContractStatus != Terminated
      .filter { it.hasMasterInception }
      .minBy { it.masterInception }.masterInception

    val periodToCheck = max(firstMasterInception?.let {YearMonth.from(it)), YearMonth.now())
```

FredrikAreschoug: To answer `What would happen if we just did "YearMonth.now()" straight up?` That's where the bug is if I have an insurance that will start in December I will get no bundle discount back from `getBundleDiscountOrNullForPeriod` since I don't have any contracts today.

FredrikAreschoug: Using Kotlins `max` would be nice but there is no implementation of it for YearMonth which would mean that I would have to get a unix timestamp of the YearMonth and then convert it back to a YearMonth when done comparing

product-pricing-service/src/main/java/com/hedvig/contract/graphql/resolvers/ContractQuery.kt
```diff
@@ -40,7 +40,18 @@ class ContractQuery(
   fun getInsuranceCost(env: DataFetchingEnvironment): InsuranceCost? {
     val memberId = env.getToken()
     val totalCost = premiumService.getTotalPremiumOfContractsWithActiveActiveInFutureOrPendingAgreementsByHolderMemberId(memberId)
-    val bundleDiscount = bundleDiscountService.getBundleDiscountOrNullForPeriod(memberId, YearMonth.now(), true)
+    //fixme: This is a lot of code in the wrong place but lets fix that by breaking out graphQL from contracts and campaign to it own module
+    val firstMasterInception = contractService.getContractStatesByHolderMemberId(memberId)
+      .filter { !it.isTerminated }
```
vonElfvin, line 45: Unfortunately this is not sufficent (agreements is a bastard really). Having a termination date is not enough to know if it is terminated today. I think this is more correct however.
```suggestion
      .filter { !it.getContractStatus != ContractStatus.TERMINATED}
```

We should probably rename "isTerminated" to "hasTerminationDate" instead

FredrikAreschoug: Okay good to know!

PR #635, opened 2020-10-14
Deprecate all product endpoints to flush out usages

vonElfvin:
I think it's time we get rid of products now. With this we'll be able to see if any other service uses old endpoints that should no longer be used. I'll spend some time hunting down those usages and make sure they use contracts instead. And then we'll be able to finally kill products once and for all.


AlexSJ1990 (approved):
Sweeeeet 🎉

PR #642, opened 2020-10-22
Feature/hvg 161 add carrier to agrement

FredrikAreschoug:
Please be thorough when looking at the upcaster and the name of the agreement value I double checked it 5 times but would be nice with an extra pair of eyes on it :slightly_smiling_face:. Spent most of the time to create a test that failed on the carrier being an enum. 

PS. fixing bugs in the deserialize is fun and all but I think we can live without the test since it worked fine when I tested it with a string in the test


product-pricing-service/src/main/java/com/hedvig/contract/shared/events/upcasters/AgreementAddedWithPredecessorAdjustedUpcaster.kt
```diff
@@ -0,0 +1,30 @@
...
+    return intermediateRepresentation.type == targetType
+  }
+
+  override fun doUpcast(intermediateRepresentation: IntermediateEventRepresentation): IntermediateEventRepresentation {
+    return intermediateRepresentation.upcastPayload(
+      SimpleSerializedType(targetType.name, "1.0"),
+      org.dom4j.Document::class.java
+    ) { document ->
+      val agreement = document.rootElement.element("agreement")
+      agreement.addElement("carrier").data = Carrier.HDI.toString()
```
vonElfvin, line 20:
```suggestion
      agreement.addElement("carrier").data = Carrier.HDI.name
```

product-pricing-service/src/main/java/com/hedvig/contract/shared/models/AgreementState.kt
```diff
@@ -54,7 +57,10 @@ abstract class AgreementState(
   open var replacedBy: AgreementState? = null,
 
   open val createdAt: Instant,
-  open var removedAt: Instant? = null
+  open var removedAt: Instant? = null,
+
+  @Enumerated(EnumType.STRING)
+  override val carrier: Carrier
```
vonElfvin, line 63: We will have to update the database manually for all the different agreements. Or, rerun all the events (which is a bit scary)

vonElfvin (approved):
Looks really good!

product-pricing-service/src/main/java/com/hedvig/contract/shared/events/upcasters/AgreementAddedWithPredecessorAdjustedUpcaster.kt
FredrikAreschoug: good catch schould really be `Carrier.HDI` the to string part was for testing :slightly_smiling_face:

product-pricing-service/src/main/java/com/hedvig/contract/shared/models/AgreementState.kt
FredrikAreschoug: Added ccolumnDefinition to prevent the scary event rerun :smile:

PR #644, opened 2020-10-23
Add ability to filter contracts of different statuses for hasContract

vonElfvin:
Hide whitespace changes 😇


product-pricing-service/src/main/java/com/hedvig/productPricing/campaign/util/DateUtil.kt
```diff
@@ -9,50 +9,54 @@ import java.time.ZoneId
 import java.time.ZoneOffset
 
 class DateUtil {
-  companion object {
```
FredrikAreschoug, line 12: I know this it really not a part of your PR but why is this a class and then everything is in a companion object couldn't a `DateUtil` just be an object?

product-pricing-service/src/main/java/com/hedvig/contract/web/InternalContractController.kt
```diff
@@ -31,125 +32,128 @@ import java.util.UUID
...
-    )
-  }
 
-  @GetMapping("/members/{memberId}/hasContract")
-  fun hasContract(@PathVariable memberId: String): ResponseEntity<Boolean> =
-    ResponseEntity.ok(contractService.hasContract(memberId))
+    @GetMapping("/members/{memberId}/hasContract")
+    fun hasContract(
+        @PathVariable memberId: String,
+        @RequestParam(required = false) statusFilter: ContractStatusFilter = ContractStatusFilter.ANY
```
FredrikAreschoug, line 152: I like it!

FredrikAreschoug (commented):
I like it my only suggestion is that It should really be `ContractStateFilter` and not status. But since But `ContractStatus` is the one that is wrong in the first place. So either or

product-pricing-service/src/main/java/com/hedvig/productPricing/campaign/util/DateUtil.kt
vonElfvin: Good catch, I changed it :)

product-pricing-service/src/main/java/com/hedvig/contract/web/InternalContractController.kt
vonElfvin: Awesome!

FredrikAreschoug (approved):
:tada:

PR #645, opened 2020-10-26
re-add deprecated endpoint is needed for approval tool

AlexSJ1990:


vonElfvin (approved):
Good catch! Didn't check the approval tool. We really should fix this later though so it doesn't use this endpoint 😄 But get this out soon so we can start doing payments.

PR #650, opened 2020-10-30
Add typeOfContractName instead so we can stop having the enum in back-office

vonElfvin:


FredrikAreschoug:
Why do we want this? cant you just change this line https://github.com/HedvigInsurance/back-office/blob/c4490946f3fed9617ff0f8ac68403bd63f4c685f/src/main/java/com/hedvig/backoffice/services/product_pricing/dto/contract/Contract.kt#L17 to a string?

PR #652, opened 2020-11-03
Feature/change number of days member is active before being charged

AlexSJ1990:
send the contractMasterInception date to account-service when creating an account -> links to pr https://github.com/HedvigInsurance/account-service/pull/104

this will allow us to check that at least a week has passed since a contract has become active before we attempt to charge a member for the first time.


FredrikAreschoug (approved):

product-pricing-service/src/main/java/com/hedvig/productPricing/service/serviceIntegration/accountService/AccountEventListener.kt
```diff
@@ -52,7 +53,7 @@ class AccountEventListener @Autowired constructor(
       logger.info("Did not send create account request since $memberId is blacklisted")
       return
     }
-    accountService.createAccount(memberId)
+    accountService.createAccount(memberId, event.activeFrom.atZone(ZoneId.of("Europe/Stockholm")).toLocalDate())
```
FredrikAreschoug, line 56: why do we only do the time zone stuff here?

